PR #6, opened 2020-07-01
Mutualize authentication

ChamseddineBhd:
Signed-off-by: chamseddine <chamseddine.benhamed@ensi-uma.tn>


src/components/AuthenticationRouter/index.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import React from 'react'
```
jonenst, line 8: is this needed ?

jonenst:
Why all the index.js files ? Are they really needed ?

src/components/AuthenticationRouter/index.js
ChamseddineBhd: no, i'll remove it

PR #13, opened 2020-08-27
Add link for merge app in top bar

med-hedi:


mikatheo:
Could we see "GridStudy" and GridMerge" appear instead of "GridStudyApp" or "GridMergeApp", with the same style as the name of the application in the top right corner (so without the hypertext link).
image

demo/src/app.js
```diff
@@ -72,7 +73,9 @@ const AppContent = () => {
                             onParametersClick={() => console.log("settings")}
                             onLogoutClick={() =>  logout(dispatch, userManager.instance)}
                             onLogoClick={() => console.log("logo")}
-                            user={user} />
+                            user={user}
+                            studyAppColor={appConfig.studyAppColor}
```
geofjamg, line 77: Why do you pass this to the top bar component, we are not going to add a new property each time we have a new supported app?

geofjamg (changes requested):

jonenst:
I think the apps and their colors should be downloaded dynamically from some known place. This way we don't have to rerelease the commons-ui components and then rerelease all the apps when we add a new app or change a color. Also all the apps will update at the same time this way

med-hedi:
@geofjamg @jonenst if i understand correctly, i should not touch the common-ui but i have to define a color code in each grid merge and grid study application

PR #27, opened 2020-09-23
Fix MaterialUI Menu error when FullScreenSUpported is false. Improve test.

jonenst:
'Material-UI: The Menu component doesn't accept a Fragment as a child.'

Signed-off-by: Jon Harper <jon.harper87@gmail.com>

### DIFF
--- a/src/components/TopBar/TopBar.test.js
+++ b/src/components/TopBar/TopBar.test.js
@@ -4,8 +4,10 @@ import expect from 'expect';
 import React from 'react';
 import { render, unmountComponentAtNode } from 'react-dom';
 import { act } from 'react-dom/test-utils';
+import { IntlProvider } from 'react-intl';
 
 import TopBar from './TopBar';
+import { top_bar_en } from '../../';
 
 import PowsyblLogo from '../images/powsybl_logo.svg';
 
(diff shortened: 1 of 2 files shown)


src/components/TopBar/TopBar.test.js
```diff
@@ -24,17 +26,19 @@ afterEach(() => {
...
+                    onLogoClick={() => {}}
+                    user={{ profile: { name: 'Username' } }}
+                />
+            </IntlProvider>,
             container
         );
     });
     expect(container.textContent).toBe(
-        '.powsybl_logo_svg__st0{fill:#ffb259}.powsybl_logo_svg__st1{fill:#6f2277}.powsybl_logo_svg__st2{fill:#fd495c}GridDemoApp'
+        '.powsybl_logo_svg__st0{fill:#ffb259}.powsybl_logo_svg__st1{fill:#6f2277}.powsybl_logo_svg__st2{fill:#fd495c}GridDemoAppUsername'
```
ChamseddineBhd, line 42: why there's no space between GridDemoApp & Username  ?

jonenst: because the .textContent gives you only the text. in the dom, they are in different elements so they are not in a single word.

ChamseddineBhd (approved):

PR #28, opened 2020-10-01
Pass apps urls and names as TopBar prop

ChamseddineBhd:
Signed-off-by: chamseddine <chamseddine.benhamed@ensi-uma.tn>


med-hedi (approved):

geofjamg (changes requested):
As logos will be different for each of the application (color) and different for light and dark theme we should be able to set the logo and logos of app list to the toolbar

geofjamg (approved):

PR #32, opened 2020-10-21
Add new component snackbar

med-hedi:
Signed-off-by: Mohamed Hedi Khefacha <mohamed.hedi@rte-france.com>


jonenst:
Can you add a button in the demo that display the message when clicked, instead of displaying it at page load ?
I don't understand how it's possible to do that with the current code.

jonenst:
Also please leave out the version from this PR. We merge PRs that bring functional changes, and we cut releases and change the version in separate PRs

med-hedi:
> Pouvez-vous ajouter un bouton dans la démo qui affiche le message lorsque vous cliquez dessus, au lieu de l'afficher au chargement de la page?
> Je ne comprends pas comment il est possible de faire cela avec le code actuel.

The idea of snackbar is to display API catch errors so there will not be buttons to click and show the message, the idea of having a notification system that appears and disappears after a short time or you can close via the closing cross, but for the demo ok I will add the button

med-hedi:
> Can you add a button in the demo that display the message when clicked, instead of displaying it at page load ?
> I don't understand how it's possible to do that with the current code.

The idea of snackbar is to display API catch errors so there will not be buttons to click and show the message, the idea of having a notification system that appears and disappears after a short time or you can close via the closing cross, but for the demo ok I will add the button

med-hedi:
> Also please leave out the version from this PR. We merge PRs that bring functional changes, and we cut releases and change the version in separate PRs

Before incrementing the version I asked the question to Chamseddine and he said yes you can do that in the same PR, but since the two must be separated, ok I do it in a new PR

jonenst:
The point of having a button is the demo is to demonstrate how to use your API when an event happens (could be a button, or an error during some asynchronous computation like you said). I don't see how you can do that with your current API. I think you will have to change the API to be able to implement the demo button.

jonenst:
Also the maxSnack value should be exposed in the API

jonenst:
I think the api is still not flexible enough to be used in a real app. Imagine that any part of the app can send a notification, how do you do that with the toplevel props "variant" and "message".
To get a sense of this, please try to have 2 buttons in the demo that show a different message when clicked.

Also hints that things can be improved:
- your <Snackbar> components is just a wrapper of a <SnackbarProvider>. This doesn't seem right, it's not the same thing.
- your <Notification> component always render an empty <Paper>

jonenst:
The current design is better but still not good:
- when clicking twice on a button in a row, only the first click shows a notification
- every time the App component renders, it displays the previous notification (this is visible for example when you modify code while using hot reload from npm start; but if you had other state in App that needed to change and cause App to rerender, it would do the same)
- when clicking a button you get the following warning in the console:
```
Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.
    in SnackBar (at app.js:187)
    in SnackbarProvider (at app.js:173)
    in ThemeProvider (at app.js:131)
    in IntlProvider (at app.js:130)
    in AppContent (at app.js:200)
    in Router (created by BrowserRouter)
    in BrowserRouter (at app.js:199)
    in App (at src/index.js:13)
```
- with the current design, if it's a deeply nested component that want to display a notification, you have to pass down the setMessage and setVariant setters through all the component hieararchy, which is bad.

Hint: why do you wrap useSnackbar() in an empty component ?

med-hedi:
@jonenst ok for the development that you do to separate the component and the call via button on the other hand I still do not see how we can integrate it into our different applications without going through the button click. I take the example of display errors in the catch of API call and if I did the same thing that you did in app.js and I add <SnackbarProvider hideIconVariant = {false}> I will have a button and not a notification.
for me the goal of adding a component in common-ui is to make it as simple as possible, in our case the call must be done via a snackbar component with variant and message props, the rest must be manage in the component itself.

jonenst:
> I take the example of display errors in the catch of API call and if I did the same thing that you did in app.js and I add I will have a button and not a notification.

No you will have only the notification. you need to:
- add SnackbarProvider (once) at the root level next to Theme and i18n Provider
- in the component launching the http request, call useSnackbar to get the enqueueSnackbar function
- in the .catch of the promise, call enqueueSnackbar

No buttons anywhere

demo/src/app.js
```diff
@@ -92,36 +137,43 @@ const AppContent = () => {
...
-                        </Typography>
-                    </Box>
-                ) : (
-                    <AuthenticationRouter
-                        userManager={userManager}
-                        signInCallbackError={null}
-                        dispatch={dispatch}
-                        history={history}
-                        location={location}
+                <SnackbarProvider hideIconVariant={false}>
```
med-hedi, line 140: I think we must add the maxSnack to define the maximum number of notifications to display

med-hedi:
@jonenst I think we can merge the PR after testing this version on gridmerge

jonenst (approved):

PR #39, opened 2020-12-08
change to react router 6

JacquesBorsenberger:


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@gridsuite/commons-ui",
-    "version": "0.5.2",
+    "version": "0.5.3",
```
ChamseddineBhd, line 3: do not change version (we change versions only for release)

ChamseddineBhd (approved):

demo/src/app.js
```diff
@@ -88,10 +87,7 @@ const AppContent = () => {
 
     // Can't use lazy initializer because useRouteMatch is a hook
     const [initialMatchSilentRenewCallbackUrl] = useState(
-        useRouteMatch({
-            path: '/silent-renew-callback',
-            exact: true,
-        })
+        location.pathname === '/silent-renew-callback'
```
jonenst, line 90: You can't use location.pathname directly because it contains the basename. Using the react router api allows to abstract it away because it's defined in the root Router

jonenst (changes requested):
react router 6 alpha has a bug, it ignores the basename. Need to wait until upstream fixes the bug and use the newer version

jonenst:
> react router 6 alpha has a bug, it ignores the basename. Need to wait until upstream fixes the bug and use the newer version

I think this was fixed since then, but can't find the issue right now

jonenst:
done in cdcaada33c0f0f1e5ff19f8949f3b7a7305f30ac

PR #51, opened 2021-04-12
Disable parameters button when callback is undefined

etiennehomer:
Signed-off-by: HOMER Etienne <etiennehomer@gmail.com>

### DIFF
--- a/src/components/TopBar/TopBar.js
+++ b/src/components/TopBar/TopBar.js
@@ -450,7 +450,7 @@ const TopBar = ({
                                         </StyledMenuItem>
 
                                         {/*/!* Equipment labelling *!/*/}
-                                        {/*If the callback onEquipmentLabellingClick is undefined, equipment labelling component isn't displayed*/}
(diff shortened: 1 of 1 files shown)


src/components/TopBar/TopBar.js
```diff
@@ -516,7 +516,9 @@ const TopBar = ({
                                         )}
 
                                         {/* Settings */}
+                                        {/*If the callback onParametersClicked is not defined, parameters component is disable*/}
```
FranckLecuyer, line 519: disabled

FranckLecuyer (changes requested):

src/components/TopBar/TopBar.js
ChamseddineBhd: If the callback onParametersClicked is **undefined**, parameters component should be disabled

ChamseddineBhd (approved):

src/components/TopBar/TopBar.js
etiennehomer: Done for both requests

FranckLecuyer (approved):

PR #76, opened 2021-09-08
remove pin of is-equal to 1.6.1, the author fixed the problem

jonenst:
https://github.com/inspect-js/is-equal/issues/77


laurgarn (approved):
A small deletion in package.json gives a lot of addition in package-lock.json. Not so intuitive ;-)

sBouzols (approved):

PR #83, opened 2021-10-05
create utility for directories element

JacquesBorsenberger:


package.json
```diff
@@ -27,7 +27,8 @@
         "prop-types": "15.7.2",
         "react-csv-downloader": "^2.7.0",
         "react-request-fullscreen": "^1.1.2",
-        "react-virtualized": "^9.21.2"
+        "react-virtualized": "^9.21.2",
+        "@material-ui/icons": "^4.5.1"
```
sBouzols, line 31: is it necessary ?

sBouzols: because it's presents in the peerdependencies List

sBouzols (approved):

package.json
JacquesBorsenberger: removed

PR #96, opened 2021-11-02
 Adapt (fit) the width of the combo list to the content (for long id or name without wrapping)

SlimaneAmar:


src/components/ElementSearchDialog/element-search-dialog.js
```diff
@@ -30,8 +31,17 @@ const useStyles = makeStyles({
         padding: '20px',
         paddingRight: '0px',
     },
+    popper: {
+        maxWidth: 'fit-content',
+    },
 });
 
+const useStyles = makeStyles(styles);
```
sBouzols, line 39: to avoid this.

src/utils/EquipmentType.js
```diff
@@ -8,9 +8,9 @@ import match from 'autosuggest-highlight/match';
 import parse from 'autosuggest-highlight/parse';
 import clsx from 'clsx';
 import React from 'react';
+import { LIGHT_THEME } from '../components/TopBar/TopBar';
```
sBouzols, line 11: Not sure we should create  direct dependency to TopBar component code here.
Maybe use :    
`import { LIGHT_THEME } from '../..';`    
which is a dependency to the commons-ui exports index (yes I know its exports from TopBar components but, you know what I mean...)
Actually we will adapt this later...

jonenst (approved):

sBouzols (approved):
Approved but some comments
demo test (light and dark theme) OK
console outputs OK

src/utils/EquipmentType.js
sBouzols: avoid import from '../..', it has terrible impacts over console warnings... don't know why.

PR #105, opened 2021-11-12
Change element type for contingencies list and filters list

FranckLecuyer:


src/utils/ElementType.js
```diff
@@ -7,31 +7,23 @@
...
 
 export const elementType = {
     DIRECTORY: 'DIRECTORY',
     STUDY: 'STUDY',
     FILTER: 'FILTER',
-    SCRIPT: 'SCRIPT',
-    SCRIPT_CONTINGENCY_LIST: 'SCRIPT_CONTINGENCY_LIST',
-    FILTERS_CONTINGENCY_LIST: 'FILTERS_CONTINGENCY_LIST',
+    CONTINGENCY_LIST: 'CONTINGENCY_LIST',
 };
```
sBouzols, line 17: We should add subtypes enums here too for CONTINGENCY_LIST and FILTER, then edit the getFileIcon function to accept a subtype parameter and keep our FilterIcon and PanToolIcon otherwise we will have a regression in gridExplore making it difficult to distinguish filter or script contingency lists and same with filters.

sBouzols (changes requested):
Do we accept a regression over element's icon in Grid Explore ?

src/utils/ElementType.js
FranckLecuyer: gridExplore dont use the getFileIcon function in commons-ui : it already selects the right icons based on the elements type and subtype.
getFileIcon is only used by study-app in directory-item-selector.js file (in directory2Tree callback), from requests to the directory-server which knows only the elements type, and not their subtypes ...

sBouzols: Ok then this getFileIcon logic by type (and subtypes) is duplicated, we should fix this later.
Anyway if we don't need the distinction in the AS it's ok for now.

sBouzols (approved):

PR #107, opened 2021-11-17
use elipsis for long ids

JacquesBorsenberger:


src/utils/EquipmentType.js
```diff
@@ -205,19 +220,18 @@ export const renderEquipmentForSearchBar = (classes, intl) => {
...
-                                    classes.equipmentTag,
-                                    classes.equipmentVlTag
-                                )}
-                            >
-                                {element.voltageLevelLabel}
-                            </span>
-                        )}
+                    </OverflowableText>
                 </div>
+                {element.type !== EQUIPMENT_TYPE.SUBSTATION.name &&
```
sBouzols, line 225: You don't need a parent `div` here with `style={{overflow: 'hidden'}}` ?
This is because you have a `equipmentVlTag` class which set a static width, then the {{overflow: 'hidden'}} inside OverflowableText works ?

sBouzols (changes requested):
There is a console warning.

src/utils/EquipmentType.js
```diff
@@ -9,8 +9,10 @@ import parse from 'autosuggest-highlight/parse';
 import clsx from 'clsx';
 import React from 'react';
 import { LIGHT_THEME } from '../components/TopBar/TopBar';
+import OverflowableText from '../components/OverflowableText';
 
 export const TYPE_TAG_MAX_SIZE = '90px';
```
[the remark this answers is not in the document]
SlimaneAmar: I do not think so
the type tag and the vl tag are fixed and the label size  in the middle take the remaining size

src/utils/EquipmentType.js
```diff
@@ -35,9 +37,14 @@ export const equipmentStyles = (theme) => ({
         background: 'lightblue',
     },
     equipmentVlTag: {
+        width: VL_TAG_MAX_SIZE,
```
SlimaneAmar, line 40: Why is not `maxWidth`

src/utils/EquipmentType.js
SlimaneAmar: Yes you can remove the div element

src/components/OverflowableText/overflowable-text.js
```diff
@@ -0,0 +1,67 @@
...
+                children={children || text}
+                className={clsx(className, classes.overflow)}
+            />
+        </Tooltip>
+    );
+};
+
+OverflowableText.propTypes = {
+    children: PropTypes.object,
+    text: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
```
SlimaneAmar, line 64: text as number ?
need to add `className` ?

src/utils/EquipmentType.js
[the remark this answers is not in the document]
JacquesBorsenberger: tag sould always have the same size so we fave a clean display

src/utils/EquipmentType.js
JacquesBorsenberger: done

src/components/OverflowableText/overflowable-text.js
JacquesBorsenberger: className added

src/utils/EquipmentType.js
```diff
@@ -191,8 +198,16 @@ export const renderEquipmentForSearchBar = (classes, intl) => {
...
                 </span>
-                <div className={classes.equipmentOption}>
-                    <span>
+                <div
+                    className={classes.equipmentOption}
+                    style={{
+                        overflow: 'hidden',
+                    }}
+                >
+                    <OverflowableText
```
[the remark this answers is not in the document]
JacquesBorsenberger: done

src/utils/EquipmentType.js
JacquesBorsenberger: done

src/components/OverflowableText/overflowable-text.js
JacquesBorsenberger: we can have have number

SlimaneAmar (approved):

src/utils/EquipmentType.js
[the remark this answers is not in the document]
sBouzols: Why type tag and the vl tag must have a fixed size ? Imagine showing this on a really wide screen ? 90px will be very small.
I thought we needed to align them vertically, reason why we have fixed size now, and we should not break responsive design as well.
We could achieve this with flexboxes and with Grid from material UI.
But anyway, this discussion is out of this PR scope.

sBouzols (approved):
Code review OK
demo tests OK
no warnings in console

PR #125, opened 2022-03-30
refactor search element dialog to permit specific tag renderer

JacquesBorsenberger:


src/utils/EquipmentType.js
```diff
@@ -172,56 +173,65 @@ const sortEquipments = (a, b) => {
...
+                    props.classes.equipmentTag,
+                    props.classes.equipmentVlTag
+                )}
+            />
         );
-    };
+    return <></>;
+};
+
+export const RenderEquipmentForSearchBar = ({
```
sBouzols, line 193: Rename by `EquipmentItem` maybe ? wich is a component now.

src/utils/EquipmentType.js
```diff
@@ -10,6 +10,7 @@ import clsx from 'clsx';
 import React from 'react';
```
sBouzols, line 10: Move this file from src/utils to src/components or src/components/ElementSearchDialog/ because its a component now not a helper function.

src/utils/EquipmentType.js
```diff
@@ -172,56 +173,65 @@ const sortEquipments = (a, b) => {
...
+                )}
+            />
         );
-    };
+    return <></>;
+};
+
+export const RenderEquipmentForSearchBar = ({
+    inputValue,
+    tagRenderer = TagRenderer,
```
sBouzols, line 195: We could use a TagRenderer by default but maybe we should use a more generic name for the component prop.
tagRenderer -> suffixRenderer ? I don't know.
It seems like this is a very generic component with 3 parts. Each part could have a proper renderer but mainly contains text or buttons. It's a pity to make it too specialized I think.

src/components/ElementSearchDialog/element-search-dialog.js
```diff
@@ -105,7 +105,14 @@ const ElementSearchDialog = (props) => {
                     noOptionsText={intl.formatMessage({
                         id: 'element_search/noResult',
                     })}
-                    renderOption={renderElement}
+                    renderOption={(prop, element, { inputValue }) =>
+                        renderElement({
+                            props,
+                            element,
+                            inputValue,
+                            close: onClose,
```
sBouzols, line 113: Why do you have to pass the onClose callback here ?

sBouzols (changes requested):
Move the component in component folder at least.

src/components/ElementSearchDialog/element-search-dialog.js
JacquesBorsenberger: because I need it ;)
in study app, to close the dialog when clicked

src/utils/EquipmentType.js
JacquesBorsenberger: done

src/utils/EquipmentType.js
JacquesBorsenberger: done

src/components/ElementSearchDialog/tag-renderer.js
```diff
@@ -0,0 +1,33 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+import OverflowableText from '../OverflowableText';
+import clsx from 'clsx';
+import React from 'react';
+import { EQUIPMENT_TYPE } from '../../utils/EquipmentType';
+import PropTypes from 'prop-types';
+
+export const TagRenderer = ({ props, element }) => {
+    if (
```
laurgarn, line 14: Cosmetic : I would have inverted the test, desindenting the bigger block.

src/utils/EquipmentType.js
```diff
@@ -172,56 +174,48 @@ const sortEquipments = (a, b) => {
...
-            </li>
-        );
-    };
+export const EquipmentItem = ({
+    inputValue,
+    suffixRenderer = TagRenderer,
+    element,
+    ...props
+}) => {
+    let matches = match(element.label, inputValue, {
```
laurgarn, line 183: Too bad,'let' is shorter than 'const' ;-)

laurgarn (approved):

sBouzols (approved):
Code Review ok
Tests demo app OK
No console warnings

PR #133, opened 2022-04-13
Fix dialog right click behaviour

Meklo:
Added a stopPropagation to the onContextMenu and onMouseDown properties of the TreeViewFinder component to prevent right-click events from being passed back to the parent components.
This will not have a negative effect as we do not generally want a right click from a dialog box to be propagated to its parents


src/utils/UIconstants.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
```
JacquesBorsenberger, line 2: 2022

JacquesBorsenberger (changes requested):
change the date for the new file ;)

PR #173, opened 2022-08-01
Read only mode for search dialog (for tree node not built)

FranckLecuyer:


src/components/ElementSearchDialog/element-search-dialog.js
```diff
@@ -66,10 +68,16 @@ const ElementSearchDialog = (props) => {
...
-                        handleSearchTermChange(value)
-                    }
-                    onChange={(_event, newValue) => onSelectionChange(newValue)}
+                    onInputChange={(_event, value) => {
+                        if (!searchTermDisabled) {
+                            handleSearchTermChange(value);
+                        }
+                    }}
+                    onChange={(_event, newValue) => {
+                        if (!searchTermDisabled && newValue) {
```
SlimaneAmar, line 77: Really needed test ?

FranckLecuyer: Yes. otherwise, there is a crash when the user deletes all characters in the textfield (newValue is null)

laurgarn (approved):

PR #174, opened 2022-08-02
keep original text with whitespaces when using OverflowableText

bendaamerahm:


FranckLecuyer (approved):
Tests: OK
Code review: OK

PR #177, opened 2022-08-25
Tamper with the demo id_token a little bit to fix the demo

jonenst:
exp 1585335149 -> 2585335149
( 2020-03-27 -> 2051-12-04 )

The demo token is a token generated by azure, the public key to verify it is
no longer available anyway. The demo can't and doesn't verify the integrity.

If we want, we could generate a token ourselves and put the public keys in this
project. If it's hardcoded it would force to serve always from the same url (issuer)
(for example localhost:3000). If needed it could also be generated on the fly depending
on the current url.. But for now the demo doesn't do anything related to token validation

Signed-off-by: HARPER Jon <jon.harper87@gmail.com>


sBouzols (approved):
Code Review OK
Test Done OK

bendaamerahm (approved):

PR #178, opened 2022-08-29
upgrade all the things!

jonenst:
babel/helper-builder-react-jsx held back because of https://github.com/babel/babel/pull/14886,
can be upgraded once the fix is released

svgr needs to stay @v5: https://github.com/gregberge/svgr/issues/665#issuecomment-1019428425
"We are moving forward, not backward. If you use webpack 4, then you should stick to SVGR v5."

eslint-loader -> eslint-webpack-plugin @ v2
from https://www.npmjs.com/package/eslint-loader "This loader has been deprecated. Please use eslint-webpack-plugin"
from https://webpack.js.org/plugins/eslint-webpack-plugin/ "This is eslint-webpack-plugin 3.0 which works only with webpack 5. For the webpack 4, see the 2.x branch."

TODO: still need to do
  "babel-eslint": "^10.1.0", -> "@babel/eslint-parser": "^7.0.0" + "@babel/eslint-plugin": "^7.0.0",
  eslint 7 -> 8
didn't work easily for me..


sBouzols (approved):
Code review ok
Demo prog tested ok
console warning heck ok

PR #184, opened 2022-10-06
Adds the missing STRING implementation for the study export options.

EstherDarkish:


dbraquart:
Code review : RAS
Tests : 
- that's OK. for example,  I can see the combo box to choose between SSH et SV while creating a study from a CGMES case.
- But I think the behaviour is very strange for other string fields, like id-mapping-file-path, boundary-location, etc. They was editable before, and now we have a "No options" combobox and these fields are no more editable. Is that what we want ?

EstherDarkish:
@dbraquart Yes, you are right, this was a bug. Fixed.
Thank you

dbraquart (approved):

PR #185, opened 2022-10-12
Highlight and expand corresponding reports when clicking on a log table row

niconoir:


src/components/ReportViewer/log-table.js
```diff
@@ -84,15 +86,36 @@ const LogTable = ({ logs }) => {
...
+    const handleRowClick = (data) => {
+        setSelectedRowIndex(data.index);
+        onRowClick(data);
+    };
+
+    const rowStyleFormat = (row) => {
+        if (row.index < 0) return;
+        if (selectedRowIndex === row.index) {
+            return {
+                backgroundColor: 'lightblue',
```
laurgarn, line 103: Does it play well with both dark and light themes ?

jonenst: the answer is no: 
Capture d’écran de 2022-11-10 10-35-04

niconoir: Modified to use colors defined for selected rows in themes.

laurgarn (approved):
Works

jonenst (approved):

PR #193, opened 2022-11-02
Add an error management when user validation is unavailable

sBouzols:
Signed-off-by: sBouzols <sylvain.bouzols@gmail.com>


jonenst (approved):

archendel (changes requested):
Works well except if user refresh pages when the new error message is displayed. In this case, an empty page appears

sBouzols:
> Works well except if user refreshes the page when the new error message is displayed. In this case, an empty page appears

Damn...

archendel (approved):
Test: OK
Code: OK

PR #199, opened 2022-11-10
use theme value to make a prettier demo table in dark mode

jonenst:
edit:  switched background from palette.primary to palette.info ( because    It's unreadable when using color: text.primary on backgroundcolor: primary.light on dark theme, was the case when hovering for the header and nonclickable text)

light:
case | before             |  after
:-------------------------:|:-------------------------:|:-------------------------:
mouse hovering on beforelast row | beforethemelight_hoverdark | afterthemelight_hoverdark
mouse hovering on last row | beforethemelight_hoverlight | afterthemelight_hoverlight

dark:
case | before | after
:-------------------------:|:-------------------------:|:-------------------------:
mouse hovering on beforelast row | beforethemedark_hoverdark | afterthemedark_hoverdark
mouse hovering on last row | beforethemedark_hoverlight | afterthemedark_hoverlight

### DIFF
--- a/demo/src/app.js
+++ b/demo/src/app.js
@@ -354,7 +354,12 @@ const AppContent = ({ language, onLanguageClick }) => {
 
     const rows = [
         { key1: 'row1_val1', key2: 'row1_val2', key3: 'row1_val3' },
-        { key1: 'row2_val1', key2: 'row2_val2', key3: 'row2_val3' },
+        {
+            key1: 'row2_val1',
+            key2: 'row2_val2',
+            key3: 'row2_val3',
+            notClickable: true,
+        },
         { key1: 'row3_val1', key2: 'row3_val2', key3: 'row3_val3' },
         { key1: 'row4_val1', key2: 'row4_val2', key3: 'row4_val3' },
     ];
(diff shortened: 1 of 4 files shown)


bendaamerahm (approved):

demo/src/app.js
```diff
@@ -360,7 +361,12 @@ const AppContent = ({ language, onLanguageClick }) => {
             key3: 'row2_val3',
             notClickable: true,
         },
-        { key1: 'row3_val1', key2: 'row3_val2', key3: 'row3_val3' },
+        {
+            key1: 'row3_val1',
+            key2: 'row3_val2',
+            key3: 'row3_val3',
+            notClickable: true,
```
sBouzols, line 368: not in the scope of this PR

demo/src/app.js
```diff
@@ -160,17 +160,18 @@ const styles = (theme) => ({
...
-        backgroundColor: 'lightblue',
+        backgroundColor: theme.palette.primary.light,
+        color: theme.palette.primary.contrastText,
         fontWeight: 'bold',
     },
     rowBackgroundDark: {
-        backgroundColor: '#81BEF7',
+        backgroundColor: theme.palette.primary.dark,
     },
     rowBackgroundLight: {
```
sBouzols, line 173: Why do we have to classes here ? should be only one set to theme.palette.primary.main and change when switching the theme no ?

sBouzols (changes requested):

demo/src/app.js
jonenst: yes you need it to test combinations on hover and rowdark rowlight

sBouzols (approved):

bendaamerahm (approved):

PR #215, opened 2022-12-30
allow search less than three characters

Walid-Sahnoun:
Signed-off-by: walid-sahnoun <walid.sahnoun@rte-france.com>

### DIFF
--- a/src/components/ElementSearchDialog/element-search-dialog.js
+++ b/src/components/ElementSearchDialog/element-search-dialog.js
@@ -11,8 +11,6 @@ import PropTypes from 'prop-types';
 import SearchIcon from '@mui/icons-material/Search';
 import { useIntl } from 'react-intl';
 
-const TERM_MIN_SIZE_BEFORE_SEARCH = 1;
-
 const ElementSearchDialog = (props) => {
     const intl = useIntl();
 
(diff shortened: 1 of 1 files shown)


laurgarn (approved):

src/components/ElementSearchDialog/element-search-dialog.js
```diff
@@ -11,7 +11,7 @@ import PropTypes from 'prop-types';
 import SearchIcon from '@mui/icons-material/Search';
 import { useIntl } from 'react-intl';
 
-const TERM_MIN_SIZE_BEFORE_SEARCH = 3;
+const TERM_MIN_SIZE_BEFORE_SEARCH = 1;
```
SlimaneAmar, line 14: Why do you keep this constant ?
I think we can delete it

SlimaneAmar (changes requested):

src/components/ElementSearchDialog/element-search-dialog.js
Walid-Sahnoun: well i tried to removed but we can't because when we search with one character and removed there is request to the server with empty term

SlimaneAmar: you need only test if the term is empty or not

SlimaneAmar (approved):

src/components/ElementSearchDialog/element-search-dialog.js
Walid-Sahnoun: done

PR #216, opened 2023-01-04
use testing-library

jonenst:


jonenst:
not using jest-dom yet for assertions

capyq (approved):

jonenst:
@capyq this will be even more useful for renderhook

and for the nice simple "rerender" function (equivalent to calling root.render again)

PR #225, opened 2023-01-30
[WIP] Use window dimensions hook

EstherDarkish:
Hook to easily get the user's displayport/window dimension even when resizes happen.


EstherDarkish:
I found an external library that does the job. This PR is not necessary anymore.

PR #227, opened 2023-02-07
Fix virtualized table forgotten legacy filter

laurgarn:


laurgarn:
Corrections spread in other PRs

PR #236, opened 2023-02-10
Demo with only one table

laurgarn:


demo/src/TableTab.js
```diff
@@ -128,6 +128,12 @@ export const TableTab = () => {
         return ret;
     }, []);
 
+    const [isIndexerExternal, setIndexerIsExternal] = useState(true);
+    const [sortable, setSortable] = useState(true);
+    const [recreats, setRecreats] = useState(false);
```
jonenst, line 133: recreates

demo/src/TableTab.js
```diff
@@ -161,82 +167,114 @@ export const TableTab = () => {
...
+        return (
+            <Stack sx={{ margin: '1ex' }}>
+                <FormControlLabel
+                    control={
+                        <Switch
+                            checked={usesCustomStyles}
+                            onChange={() => setUsesCustomStyles((was) => !was)}
+                        />
+                    }
+                    // labelPlacement={'start'}
```
jonenst, line 185: remove instead of comment

demo/src/TableTab.js
laurgarn: done

demo/src/TableTab.js
laurgarn: done

jonenst (approved):

PR #243, opened 2023-02-23
fix: filter icon color

laurgarn:


jonenst:
val9 doesn't exist in head2 col, so the filter icon should become errored ("classes.filterTooLossy")

before:
Capture d’écran de 2023-05-24 10-44-16

after:
Capture d’écran de 2023-05-24 10-44-33

jonenst (approved):

PR #259, opened 2023-05-15
Emotion

jonenst:


demo/src/right-resizable-box.js
```diff
@@ -66,27 +68,26 @@ const RightResizableBox = (props) => {
     };
 
     return (
-        <ResizableBox
+        <ResizableBoxSx
             style={{ display: props.hide ? 'none' : undefined }}
```
AbdelHedhili, line 72: why isn't this in the sx ?

demo/src/app.js
```diff
@@ -623,6 +657,28 @@ const AppContent = ({ language, onLanguageClick }) => {
                             </div>
                             <div style={{ flexGrow: 1 }} />
                             <div style={{ alignSelf: 'center' }}>baz</div>
```
AbdelHedhili, line 659: sx ? Or you don't want to change the div to a box just for that ?

src/components/MuiVirtualizedTable/MuiVirtualizedTable.js
```diff
@@ -814,6 +852,7 @@ class MuiVirtualizedTable extends React.PureComponent {
                     flexDirection: 'column',
                     height: '100%',
                 }}
+                className={this.props.className}
             >
```
AbdelHedhili, line 856: why aren't all the static styles 'sxed' in this part

AbdelHedhili (commented):
code: done
test: in progress

demo/src/right-resizable-box.js
jonenst: I migrated makeStyle/useStyle to emotion but left everything else as is. Can be refactored in a separate PR

demo/src/app.js
jonenst: I migrated makeStyle/useStyle to emotion but left everything else as is. Can be refactored in a separate PR

src/components/MuiVirtualizedTable/MuiVirtualizedTable.js
jonenst: I migrated makeStyle/useStyle to emotion but left everything else as is. Can be refactored in a separate PR

AbdelHedhili (dismissed):
the topbar style is broken
toolbar

jonenst:
fixed margin, with sx it's not pixels when providing a number, it's multiple of theme.spacing (default 8px) https://mui.com/system/getting-started/the-sx-prop/#spacing


fixed report viewer sx was overwritten by parent instead of passing the classname

Walid-Sahnoun (commented):
logs in left side is broken
there is no horizontal scroll, then the log is written in more than 1 line.
also, the highlight is not showed when you hover on second or third level. 
image

src/utils/styles.js
```diff
@@ -0,0 +1,29 @@
...
+export const makeComposeClasses =
+    (generateGlobalClass) => (classes, ruleName) =>
+        [generateGlobalClass(ruleName), classes[ruleName]]
+            .filter((x) => x)
+            .join(' ');
+
+export const toNestedGlobalSelectors = (styles, generateGlobalClass) =>
+    Object.fromEntries(
+        Object.entries(styles).map(([k, v]) => [
+            `& .${generateGlobalClass(k)}`,
```
flomillot, line 21: Why do we have to use this ?

flomillot: I mean, we can't get rid of classNames ?

demo/src/app.js
```diff
@@ -394,6 +412,13 @@ const AppContent = ({ language, onLanguageClick }) => {
         []
     );
 
+    const CustomTreeViewFinder =
+        stylesProvider === 'emotion'
```
flomillot, line 416: Why do we keep both way ?

jonenst: I made a backwards compatible migration. It's still possible to generate classnames with jss, and then use the classes APIs to apply the classes. This is to make migrations easier. It's what mui did. So in the demo, we can test both modes. If mui removes the classes API, we will have to drop support for js, but for now it's easy to support

src/utils/styles.js
flomillot: Yes, but almost everytime, we stylise the component itself and not nested components.
Or I missed something ?

demo/src/right-resizable-box.js
```diff
@@ -38,19 +39,20 @@ const useStyles = makeStyles((theme) => ({
...
         transform: 'scale(0.5, 1.5)',
-    },
-}));
+    }),
+};
 
-const RightResizableBox = (props) => {
-    const classes = useStyles();
+// TODO can we avoid to define a component just to add sx support ?
+const ResizableBoxSx = styled(ResizableBox)({});
```
flomillot, line 53: Instead of doing this and using mergeSx, the documentation suggest to do it in an other way, did you see it ?
https://mui.com/material-ui/customization/how-to-customize/#dynamic-overrides

flomillot (approved):

package.json
```diff
@@ -39,7 +39,6 @@
         "@mui/icons-material": "^5.5.0",
         "@mui/lab": "^5.0.0-alpha.72",
         "@mui/material": "^5.5.0",
-        "@mui/styles": "^5.5.0",
```
sBouzols, line 42: remove it from devDependencies as well ?

demo/src/right-resizable-box.js
```diff
@@ -66,27 +68,26 @@ const RightResizableBox = (props) => {
...
             style={{ display: props.hide ? 'none' : undefined }}
             width={
                 props.fullscreen
                     ? windowWidth
                     : windowWidth * resizedTreePercentage
             }
-            className={clsx(classes.panel, {
-                [classes.resizePanelHandle]: !props.disableResize,
-            })}
+            sx={mergeSx(
```
sBouzols, line 78: cf doc : https://www.npmjs.com/package/react-resizable?activeTab=readme
it says : "Unknown props that are not in the list below will be passed to the child component."
nor `className` nor `sx` are a known prop, maybe set it directly to the Box child ?
Or I miss something ?

demo/src/app.js
```diff
@@ -623,6 +657,28 @@ const AppContent = ({ language, onLanguageClick }) => {
...
+                                size="small"
+                            >
+                                <InputLabel id="styles-provider-label">
+                                    Custom theme
+                                </InputLabel>
+                                <Select
+                                    labelId="styles-provider-label"
+                                    id="styles-provider"
+                                    value={stylesProvider}
+                                    label="Styles Provider"
```
sBouzols, line 671: missing translation

demo/src/app.js
```diff
@@ -623,6 +657,28 @@ const AppContent = ({ language, onLanguageClick }) => {
                             </div>
                             <div style={{ flexGrow: 1 }} />
                             <div style={{ alignSelf: 'center' }}>baz</div>
+                            <FormControl
+                                sx={{ m: 1, minWidth: 120 }}
+                                size="small"
+                            >
+                                <InputLabel id="styles-provider-label">
+                                    Custom theme
```
sBouzols, line 665: missing translation

package.json
sBouzols: I see you want to keep it for the demo app to compare ok.

src/components/OverflowableText/overflowable-text.js
```diff
@@ -22,50 +21,57 @@ const overflowStyle = (theme) => ({
         width: 'fit-content',
         maxWidth: 'fit-content',
     },
-});
+};
 
-const useStyles = makeStyles(overflowStyle);
+export const OverflowableText = styled(
```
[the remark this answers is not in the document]
jonenst: I used styled to add sx support on the component to provide an alternative to the className API

demo/src/right-resizable-box.js
jonenst: sx is handled by styled and transformed to a classname prop on the component.
Not sure what's best between setting them on the resizable or the box

src/components/MuiVirtualizedTable/ColumnHeader.js
```diff
@@ -40,52 +41,50 @@ const useStyles = makeStyles((theme) => ({
...
     );
 };
 
 const FilterButton = (props) => {
-    const classes = useStyles();
-    const visibilityClass =
+    const visibilityStyle =
         !props.filterLevel &&
-        (props.headerHovered ? classes.hovered : classes.transparent);
+        (props.headerHovered ? styles.hovered : styles.transparent);
```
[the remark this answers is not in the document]
jonenst: didn't do it because it's not necessary in line 80 wheareas it's necessary at line 63... because line 80 uses mergeSx because we want to combine multiple rules, but line 63 only has the visibility rules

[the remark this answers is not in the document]
jonenst: But it could be done

demo/src/app.js
EstherDarkish: Added.

PR #262, opened 2023-05-23
Fix ReportViewer report-item treeitem css after mui v5 migration

jonenst:
in https://github.com/mui/material-ui/pull/21514 pseudo classes were moved from root to content, so we no longer need nested selectors and rules need to go on content directly use Mui-focused instead of :focus

also restore muiv4 treeitem label color rule to get a different color right after a click on an item. It was also removed in treeitem in https://github.com/mui/material-ui/pull/21514 Note: in v5 primary.main was changed from indigo to blue in https://github.com/mui/material-ui/pull/26555) but let's go with the new primary colors
Note2: fade renamed to alpha


EstherDarkish (approved):

sBouzols (approved):
Code Review OK
Test OK
Some style to adapt, will see in another PR

PR #263, opened 2023-05-23
Add alternative to local path in README.md

trikimahdi:


SlimaneAmar (approved):

trikimahdi:
The test with local path @gridsuite/commons-ui:'file:../path/to/the/commons-ui' was successful. We don't need to add another format in README.md

PR #279, opened 2023-08-04
Add dialog to select extentions in study creation

YenguiSeddik:


src/components/MultipleSelectionDialog/MultipleSelectionDialog.js
```diff
@@ -0,0 +1,125 @@
...
+                                control={
+                                    <Checkbox
+                                        checked={
+                                            selectedIds.length ===
+                                            options.length
+                                        }
+                                        indeterminate={
+                                            selectedIds.length !==
+                                                options.length &&
+                                            selectedIds.length !== 0
```
ghelloumi, line 69: I think that is better to do

selectedIds.length && selectedIds.length !== options.length 

(it is unnecessary to test selectedIds.length !== options.length if there is no selectedIds)

src/components/FlatParameters/FlatParameters.js
```diff
@@ -268,6 +300,52 @@ export const FlatParameters = ({
                 );
             case 'STRING_LIST':
                 if (param.possibleValues) {
+                    const allOptions = sortPossibleValues(
+                        param.name,
+                        param.possibleValues
+                    ).map((v) => v.id);
```
ghelloumi, line 306: .map(({id}) => id);

src/components/FlatParameters/FlatParameters.js
```diff
@@ -211,6 +231,18 @@ export const FlatParameters = ({
         return value?.replace(',', '.') || '';
     };
 
+    const getStringListValue = (allValues, selectValues) => {
+        if (!Array.isArray(selectValues) || selectValues?.length === 0) {
```
ghelloumi, line 235: can selectedValues be with type other than array if defined?
if not you can simply use
if (!selectValues || !selectValues.length)

ghelloumi (approved):
code ok
test ok

ayolab (approved):

PR #280, opened 2023-08-10
Switch the build from ES5 using CommonsJS to ES6 using ES Modules

flomillot:


README.md
archendel, line 1: revert this change

archendel (approved):

SlimaneAmar (approved):

jonenst:
Is there a link to nwb documentation explaining this system ?
What are the technical effects of this change?
Why are we doing this change, what do we want from it ?

flomillot:
> Is there a link to nwb documentation explaining this system ? What are the technical effects of this change? Why are we doing this change, what do we want from it ?

https://github.com/insin/nwb/blob/master/docs/guides/ReactComponents.md#npm-build-configuration
The main difference is the way other modules are imported inside commons-ui.
The reason is as we switched the build to ES6 in gridstudy / gridexplore when we moved to TypeScript, if we want to use the same imports when we use commons-ui, we have to switch commons-ui to ES6 too.
Some explications in US2176.

PR #284, opened 2023-08-18
Add filtering of log severity

Meklo:


ayolab (approved):

src/components/ReportViewer/multi-select-list.js
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * MultiSelectList allows to manipulate an object where each keys are associated to a boolean in order to determine which are the ones the user wants to select
+ * A visual indication is displayed when the user alters the default state of the list
+ *
+ * @param {Object}   selectedItems    - It serves as the model and data of the Component each entry must be formatted as a pair of string and boolean. Each key will be displayed and the corresponding boolean is updated in function of its checkbox status
+ * @param {Function} setSelectedItems - Setter needed to update the list underlying data
+ */
+
+export const MultiSelectList = ({ selectedItems, setSelectedItems }) => {
```
archendel, line 47: it could be nice to dissociate the MultiSelectList, which is a component use to select multiple elements, and the filterElement, which is a clickable button, which displays the MultiSelectList when clicked

if we need another type of filter, it will be really easier to develop
and it will make both components easier to maintain / understand

src/components/ReportViewer/log-table.js
```diff
@@ -114,6 +126,18 @@ const LogTable = ({ logs, onRowClick }) => {
         setSelectedRowIndex(-1);
     }, [logs]);
 
+    const filter = useCallback(
+        (row) => {
+            return (
+                row.severity &&
+                Object.entries(selectedSeverity).some(
+                    ([key, value]) => key === row.severity && value
+                )
```
archendel, line 135: we are filtering rows which has falsy severity or false value
it was not done before, is it necessary ?

archendel (changes requested):

src/components/ReportViewer/multi-select-list.js
Meklo: Agreed, I've relocated the button part in a new component

src/components/ReportViewer/log-table.js
Meklo: After some investigation I've found some reports without severity in the db but those are dedicated to network metadata so they aren't linked to a node thus they wouldn't appear in the report viewer

src/components/ReportViewer/filter-button.js
```diff
@@ -0,0 +1,79 @@
...
+    };
+
+    const handleChange = (event) => {
+        setSelectedItems((previousSelection) => {
+            return {
+                ...previousSelection,
+                [event.target.name]: !selectedItems[event.target.name],
+            };
+        });
+    };
```
archendel, line 57: I think it should be handled by parent component, otherwise it makes it linked to current filter behaviour

src/components/ReportViewer/filter-button.js
```diff
@@ -0,0 +1,79 @@
...
+
+    const isInitialStateModified = useMemo(() => {
+        return Object.keys(selectedItems).some(
+            (key) => initialState[key] !== selectedItems[key]
+        );
+    }, [initialState, selectedItems]);
```
archendel, line 63: we might should do the same here, make it a dumb component
add a prop like "displayHasChangedNotification" to this component, and let the parent decide whether it should be displayed or not

archendel (approved):

PR #298, opened 2023-09-19
Fix Warning OverflowableText sx propTypes

sBouzols:
fix(OverflowableText): fix sx propTypes to avoid warning when using arrays (mergeSx) for the sx props.

Warning: Failed prop type: Invalid prop `sx` of type `array` supplied to `Styled(Component)`, expected `object`.


EstherDarkish (approved):

EstherDarkish (approved):
Code OK, Tests OK

PR #300, opened 2023-09-21
Fix ElementSearchDialog: Add clearOnBlur prop 

sBouzols:
Because even with value=null when in FreeSolo mode the AutoComplete remains the value in the input.


Tristan-WorkGH (approved):

antoinebhs (approved):
Code OK
Tests OK

PR #307, opened 2023-10-18
add cancel button

anistouri:


YenguiSeddik (approved):

src/components/react-hook-form/utils/cancel-button.js
```diff
@@ -0,0 +1,18 @@
...
+import { Button } from '@mui/material';
+import { FormattedMessage } from 'react-intl';
+
+const CancelButton = ({ onClick, variant, disabled }) => {
+    return (
+        <Button
+            onClick={onClick}
+            variant={variant}
+            disabled={disabled}
+            sx={{ color: 'inherit' }}
```
flomillot, line 11: Why don't you use the color of the theme as suggested in the US ? How can you be sure inherit give you the color you want ?

flomillot: What happen if you remove the inherit ?

anistouri: If I remove "inherit", all the buttons take on the default MUI color which is blue in gridexplore. That's the reason why I placed this component in commons UI, so I don't have to repeat "inherit" and it's only to cancel it out.

src/components/react-hook-form/utils/cancel-button.js
```diff
@@ -0,0 +1,28 @@
...
+    variant,
+    disabled,
+    withCustomColor = true,
+}) => {
+    return (
+        <Button
+            onClick={onClick}
+            variant={variant}
+            disabled={disabled}
+            sx={withCustomColor ? styles.removeButtonTextColor : null}
```
flomillot, line 21: Why you use `sx` and not the `color` prop ?
https://mui.com/material-ui/api/button/#Button-prop-color

src/components/react-hook-form/utils/cancel-button.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { Button } from '@mui/material';
+import { FormattedMessage } from 'react-intl';
+
+const CancelButton = ({
+    onClick,
+    variant,
+    disabled,
+    withCustomColor = false,
```
flomillot, line 9: Either rename it in "withOriginalColor", or change the condition below.

src/components/react-hook-form/utils/cancel-button.js
```diff
@@ -0,0 +1,23 @@
...
+}) => {
+    return (
+        <Button
+            onClick={onClick}
+            variant={variant}
+            disabled={disabled}
+            color={withCustomColor ? 'primary' : 'secondary'}
```
flomillot, line 16: Don't use secondary it already exist

anistouri: In the app, "secondary" refers to the pink color. I've overloaded this parameter because the name "secondary" is more descriptive than anything else, but I can change it if you have another name in mind

anistouri: I replaced it with **customButton**

flomillot (approved):

PR #321, opened 2023-11-30
Add about dialog

Tristan-WorkGH:

PR #326, opened 2023-12-14
Fix about window

Tristan-WorkGH:
Following #321

Fix "About" window:
* checked tag version is always visible, even with long tag
* fixed "moving" version when loading (due to animation and bold version not having same height than normal text)
* when closing the dialog, not having during fade-out animation value flashing/reset

### DIFF
--- a/src/components/TopBar/AboutDialog.js
+++ b/src/components/TopBar/AboutDialog.js
@@ -22,6 +22,7 @@ import {
     Fade,
     Grid,
     Stack,
+    Tooltip,
     Typography,
     useMediaQuery,
     useTheme,
@@ -358,10 +359,24 @@ AboutDialog.propTypes = {
     getAdditionalModules: PropTypes.func,
 };
 
+const style = {
+    icons: {
+        flexGrow: 0,
+        position: 'relative',
+        top: '4px',
+        flexShrink: 0,
+    },
+    version: {
+        flexGrow: 0,
+        alignSelf: 'flex-end',
+        flexShrink: 0,
+    },
+};
+
 const ModuleTypesIcons = {
-    app: <WidgetsOutlined fontSize="small" color="primary" />,
-    server: <DnsOutlined fontSize="small" color="secondary" />,
-    other: <QuestionMark fontSize="small" />,
+    app: <WidgetsOutlined sx={style.icons} fontSize="small" color="primary" />,
+    server: <DnsOutlined sx={style.icons} fontSize="small" color="secondary" />,
+    other: <QuestionMark sx={style.icons} fontSize="small" />,
 };
 
 const Module = ({ type, name, version, license }) => {
@@ -377,26 +392,28 @@ const Module = ({ type, name, version, license }) => {
                 },
             }}
         >
-            <Stack
-                direction="row"
-                justifyContent="flex-start"
-                alignItems="baseline"
-                spacing={1}
-            >
-                {ModuleTypesIcons[type] || ModuleTypesIcons['other']}
-                <Typography display="inline-block" noWrap>
-                    {name || '<?>'}
-                </Typography>
-                <Typography
-                    variant="caption"
-                    color={(theme) => theme.palette.text.secondary}
-                    display="inline"
-                    marginLeft={1}
-                    noWrap
+            <Tooltip title={(name || '<?>') + ' ' + (version || '')}>
+                <Stack
+                    direction="row"
+                    justifyContent="flex-start"
+                    alignItems="baseline"
+                    spacing={1}
                 >
-                    {version || null}
-                </Typography>
-            </Stack>
+                    {ModuleTypesIcons[type] || ModuleTypesIcons['other']}
+                    <Typography display="inline" noWrap>
+                        {name || '<?>'}
+                    </Typography>
+                    <Typography
+                        variant="caption"
+                        color={(theme) => theme.palette.text.secondary}
+                        display="inline"
+                        noWrap
+                        sx={style.version}
+                    >
+                        {version || null}
+                    </Typography>
+                </Stack>
+            </Tooltip>
         </Grid>
     );
 };
--- a/src/components/TopBar/AboutDialog.js
+++ b/src/components/TopBar/AboutDialog.js
@@ -78,8 +78,6 @@ function getGlobalVersion(fnPromise, type, setData, setLoader) {
     }
 }
 
-const dialogExitTransitionDurationMs = 195;
-
 const moduleTypeSort = {
     app: 1,
     server: 10,
@@ -106,6 +104,7 @@ const AboutDialog = ({
     const theme = useTheme();
     const [isRefreshing, setRefreshState] = useState(false);
     const [loadingGlobalVersion, setLoadingGlobalVersion] = useState(false);
+    const [showGlobalVersion, setShowGlobalVersion] = useState(false);
 
     /* We want to get the initial version once at first render to detect later a new deploy */
     const [startingGlobalVersion, setStartingGlobalVersion] =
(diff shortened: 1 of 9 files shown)


src/components/TopBar/AboutDialog.js
```diff
@@ -116,6 +118,17 @@ const AboutDialog = ({
...
                 setModules([currentApp]);
             }
+        } else {
+            // we wait the end of the fade animation of the dialog before reset content
+            setTimeout(
+                (setModules, setActualGlobalVersion) => {
+                    setModules(null);
+                    setActualGlobalVersion(null);
+                },
+                195 + 5,
```
EstherDarkish, line 128: I solved this : 200

src/components/TopBar/AboutDialog.js
```diff
@@ -392,7 +446,62 @@ const Module = ({ type, name, version, license }) => {
...
+                                        }) === 0
+                                            ? 'app'
+                                            : 'server'.localeCompare(
+                                                  type,
+                                                  undefined,
+                                                  { sensitivity: 'base' }
+                                              ) === 0
+                                            ? 'server'
+                                            : 'other'
+                                    }`}
```
EstherDarkish, line 478: This is not maintenable code. Too hard to read. Please refactor this.

EstherDarkish (changes requested):

src/components/TopBar/AboutDialog.js
```diff
@@ -392,7 +408,62 @@ const Module = ({ type, name, version, license }) => {
                 },
             }}
         >
-            <Tooltip title={(name || '<?>') + ' ' + (version || '')}>
+            <Tooltip
+                TransitionComponent={Zoom}
+                enterDelay={2500}
```
flomillot, line 413: It's so slow... 1000 maximum

src/components/TopBar/AboutDialog.js
```diff
@@ -116,6 +117,17 @@ const AboutDialog = ({
...
                 setModules([currentApp]);
             }
+        } else {
+            // we wait the end of the fade animation of the dialog before reset content
+            setTimeout(
+                (setModules, setActualGlobalVersion) => {
+                    setModules(null);
+                    setActualGlobalVersion(null);
+                },
+                195 + 5,
```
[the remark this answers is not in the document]
flomillot: If it's related to the 195 above I think you should use a const

flomillot (changes requested):

src/components/TopBar/AboutDialog.js
Tristan-WorkGH: That wanted, because it spam the view if you're juste scrolling in the list

EstherDarkish: I could see 500 by the way...

EstherDarkish: Better ask Stéphane about this

src/components/TopBar/AboutDialog.js
```diff
@@ -41,6 +43,97 @@ import { FormattedMessage } from 'react-intl';
...
+                whiteSpace: 'pre',
+            },
+            '&:first-of-type': {
+                //...but not for the first line
+                '&:before': {
+                    content: "''",
+                },
+            },
+        },
+    },
```
EstherDarkish, line 85:
```suggestion
    mainInfos: {
        //dl section contains dt (key) + dd (value) tuples
        textAlign: 'center',
        marginTop: 0,
        'dt, dd': {
            display: 'inline',
            margin: 0,
        },
        dt: {
            marginRight: '0.5em',
        },
    },
```

src/components/TopBar/AboutDialog.js
```diff
@@ -58,39 +150,46 @@ function compareModules(c1, c2) {
...
-        if (startingGlobalVersion === undefined && getGlobalVersion) {
-            getGlobalVersion((value) => {
-                setStartingGlobalVersion(value);
-                setActualGlobalVersion(value);
-            });
+        if (startingGlobalVersion === undefined && globalVersionPromise) {
+            getGlobalVersion(
+                globalVersionPromise,
+                'Initial',
+                setStartingGlobalVersion,
```
flomillot, line 173:
```suggestion
                setInitialGlobalVersion,
```
Just to have the same term as above

src/components/TopBar/AboutDialog.js
```diff
@@ -58,39 +150,46 @@ function compareModules(c1, c2) {
...
-                setLoadingGlobalVersion(false);
-                setActualGlobalVersion(value || null);
-            });
+        if (open && globalVersionPromise) {
+            getGlobalVersion(
+                globalVersionPromise,
+                'Actual',
+                setActualGlobalVersion,
+                (value) => {
+                    setLoadingGlobalVersion(value);
```
flomillot, line 187:
```suggestion
                    setLoadingGlobalVersion(!!value);
```
More comprehensive no ?

src/components/TopBar/AboutDialog.js
```diff
@@ -350,11 +424,11 @@ AboutDialog.propTypes = {
...
     appGitTag: PropTypes.string,
     appLicense: PropTypes.string,
-    getGlobalVersion: PropTypes.func,
-    getAdditionalModules: PropTypes.func,
+    globalVersionPromise: PropTypes.func,
+    additionalModulesPromise: PropTypes.func,
 };
 
-const style = {
+const ModuleStyles = {
```
flomillot, line 431: I don't think we should start by upper case except for react components

src/components/TopBar/AboutDialog.js
```diff
@@ -42,6 +43,97 @@ import { FormattedMessage } from 'react-intl';
...
+    detailsSection: {
+        '.MuiAccordionSummary-content > .MuiSvgIcon-root': {
+            marginRight: '0.5rem',
+        },
+    },
+};
+
+function getGlobalVersion(fnPromise, type, setData, setLoader) {
+    if (fnPromise) {
+        console.debug('Getting', type, 'globale version...');
```
flomillot, line 104:
```suggestion
        console.debug('Getting', type, 'global version...');
```

src/components/TopBar/AboutDialog.js
```diff
@@ -186,72 +285,47 @@ const AboutDialog = ({
...
-                                        fontSize: '1em',
-                                    }}
-                                />
+                    <Box component="dl" sx={styles.mainInfos}>
+                        <Fade
+                            in={loadingGlobalVersion}
+                            appear
+                            unmountOnExit
+                            onExited={(node) => {
+                                if (open && !loadingGlobalVersion) {
```
flomillot, line 301:
```suggestion
                                if (open) {
```
`loadingGlobalVersion` is already false no ?

src/components/TopBar/AboutDialog.js
```diff
@@ -151,12 +240,22 @@ const AboutDialog = ({
             fullWidth={true}
             maxWidth="md"
             fullScreen={useMediaQuery(theme.breakpoints.down('md'))}
+            sx={styles.general}
             aria-labelledby="alert-dialog-title"
             aria-describedby="alert-dialog-description"
-            transitionDuration={{ exit: 195 }}
+            TransitionProps={{
+                onExited: (node) => {
+                    if (!open) {
```
flomillot, line 248: Same, how can it be true ?

flomillot (commented):
As discussed on Slack, double check on promises to remove

src/components/TopBar/AboutDialog.js
Tristan-WorkGH: no, you can close the dialog while it still is loading

EstherDarkish (approved):

src/components/TopBar/AboutDialog.js
flomillot: Ok but then what's the point of intering here because we want to execute this code only when `open` is true ?

flomillot (approved):
Code OK

PR #344, opened 2024-02-06
Clean MUI imports

flomillot:
Code cleaning on imports that was initialy done on #352 because of issues with default imports.

We use only named imports from first level for MUI https://github.com/mui/material-ui/issues/30671#issuecomment-1613525947
It's strongly advised to only use named imports in libraries, cf https://esbuild.github.io/content-types/#default-interop

> If you are a library author: When writing new code, you should strongly consider avoiding the default export entirely. It has unfortunately been tainted with compatibility problems and using it will likely cause problems for your users at some point.


vite.config.ts
```diff
@@ -0,0 +1,70 @@
+import react from '@vitejs/plugin-react';
```
sBouzols, line 1: typescript file -> license header

package.json
```diff
@@ -1,27 +1,23 @@
 {
     "name": "@gridsuite/commons-ui",
-    "version": "0.43.0",
+    "version": "0.44.0",
```
klesaulnier, line 3: should we update commons-ui version in a specific PR as usual ?

src/components/TopBar/TopBar.jsx
```diff
@@ -40,8 +40,7 @@ import {
     Settings as SettingsIcon,
     WbSunny as WbSunnyIcon,
 } from '@mui/icons-material';
-import { darken } from '@mui/material/styles';
-import { styled } from '@mui/system';
+import { darken, styled } from '@mui/system';
```
klesaulnier, line 43:
```suggestion
import { styled } from '@mui/system';
import { darken } from '@mui/material';
```

tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "compilerOptions": {
+    "module": "ESNext",
```
klesaulnier, line 3:
```suggestion
    "module": "esnext",
```

package.json
flomillot: Yes mb

src/components/TopBar/TopBar.jsx
flomillot: It's actually the same

tsconfig.json
flomillot: Done

vite.config.ts
flomillot: Done

jest.config.ts
```diff
@@ -0,0 +1,13 @@
+import type { Config } from 'jest';
```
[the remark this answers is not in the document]
flomillot: Done

klesaulnier (approved):

Tristan-WorkGH:
@flomillot I've resolved the `legacy-peer-deps` in #350

Tristan-WorkGH:
@flomillot I think I've fixed the external dependencies with #351

Tristan-WorkGH (approved):

PR #350, opened 2024-02-16
Resolve legacy peers

Tristan-WorkGH:
Resolve the `legacy-peer-deps` problem which is caused currently by `@mui/styles`, which depend to React 17.
* `@mui/styles` is [officially **deprecated**](https://mui.com/system/styles/basics/) since MUI5
  * replaced by [`@mui/system`](https://mui.com/system/getting-started/), or it's variant `@mui/material/styled`
    * use by default emotion
* the old JSS system is only used in the demo with `TableTab` and `EquipementItem` components
  * `EquipementItem` doesn't depend explicitly to JSS
    * only used by GridStudy and only used emotion system
      EDIT: has been said it was to compare the visual of components before migrating to emotion

Used [migration guide](https://mui.com/material-ui/migration/migrating-from-jss/) with the help of [`@mui/codemod`](https://github.com/mui/material-ui/blob/master/packages/mui-codemod/README.md#jss-to-styled).

---

Update:
Some libraries in #399 wasn't upgraded because of `legacy-peers-deps` resolution causing conflicts, but mandatory when removing it:
- bump `eslint-config-react-app` for compatibility to `@typescript-eslint/eslint-plugin`
- bump `eslint` for compatibility to `eslint-config-react-app`
- add `@babel/plugin-proposal-private-property-in-object` to hide warning during lint
  <details>
    <summary>Click me</summary>

    >One of your dependencies, babel-preset-react-app, is importing the "@babel/plugin-proposal-private-property-in-object" package without declaring it in its dependencies. This is currently working because "@babel/plugin-proposal-private-property-in-object" is already in your node_modules folder for unrelated reasons, but it may break at any time.
    >
    >babel-preset-react-app is part of the create-react-app project, which is not maintianed anymore. It is thus unlikely that this bug will ever be fixed. Add "@babel/plugin-proposal-private-property-in-object" to your devDependencies to work around this error. This will make this message go away.
  </details>


flomillot (changes requested):
This withStyles is even older than the one we use.

Cf from `node_modules/@mui/material/styles/withStyles.js`

```
import { formatMuiErrorMessage as _formatMuiErrorMessage } from "@mui/utils";
export default function withStyles() {
  throw new Error(process.env.NODE_ENV !== "production" ? `MUI: withStyles is no longer exported from @mui/material/styles.
You have to import it from @mui/styles.
See https://mui.com/r/migration-v4/#mui-material-styles for more details.` : _formatMuiErrorMessage(15));
}
```

Or I missed something ?

Actually the only clean solution is to totally remove `withStyles` and replace by the new system. But it's much more work.

Tristan-WorkGH:
> First, we should only import MUI components from first level using named imports (we had issues during vite migration, you can check the explanations in the US).
> 
> Then, this withStyles is even older than the one we use.
> 
> Cf `node_modules/@mui/material/styles/withStyles.js`
> 
> ```
> import { formatMuiErrorMessage as _formatMuiErrorMessage } from "@mui/utils";
> export default function withStyles() {
>   throw new Error(process.env.NODE_ENV !== "production" ? `MUI: withStyles is no longer exported from @mui/material/styles.
> You have to import it from @mui/styles.
> See https://mui.com/r/migration-v4/#mui-material-styles for more details.` : _formatMuiErrorMessage(15));
> }
> ```
> 
> Explanations here https://stackoverflow.com/questions/69506133/difference-between-mui-material-styles-and-mui-styles
> 
> Actually the only clean solution is to totally remove `withStyles` and replace it by the new system. But it's much more work...
> 
> A part of the job has already been done here #259
> 
> I suggest to close the PR except if you're motived to do this... ?

Oh, that was what I did before, but I was afraid it was too much for 1 PR, so I split-ted in #350 & #354.  
I will re-merge the two then.

demo/src/TableTab.jsx
```diff
@@ -18,14 +15,14 @@ import {
     Switch,
     TextField,
 } from '@mui/material';
+import { DEFAULT_CELL_PADDING, KeyedColumnsRowIndexer } from '../../src';
```
flomillot, line 18: Put it back to same place ?

demo/src/app.jsx
```diff
@@ -656,7 +639,7 @@ const AppContent = ({ language, onLanguageClick }) => {
                         setOpenTreeViewFinderDialogCustomDialog(true)
                     }
                 >
-                    Open Custom TreeViewFinder ({stylesProvider}) ...
+                    Open Custom TreeViewFinder (emotion) ...
```
flomillot, line 642: We still need to specify it's emotion ?

flomillot (commented):
Code OK (no tested)

flomillot (approved):

demo/src/app.jsx
Tristan-WorkGH: I found it logic, but dunno

Tristan-WorkGH:
There was too much diff with `main` branch with migration to Vite, so have restart from main.

Tristan-WorkGH:
waiting #399 to have updated dependencies

src/utils/EquipmentType.ts
```diff
@@ -19,12 +21,17 @@ export const equipmentStyles = {
...
     },
-    equipmentTag: (theme: string) => ({
+    equipmentTag: (theme: string | Theme) => ({
         borderRadius: '10px',
         padding: '4px',
         fontSize: 'x-small',
         textAlign: 'center',
-        color: theme === LIGHT_THEME ? 'inherit' : 'black',
+        color:
+            //TODO remove first condition when gridstudy is updated
```
flomillot, line 30: Still need this ?

flomillot (approved):
Code OK, no warning, just one comment.

src/utils/EquipmentType.ts
Tristan-WorkGH: yes some fronts like gridstudy pass a string instead on a `Theme` object...

PR #355, opened 2024-02-22
Add TypeScript support

flomillot:
Import and merge the types declarations for commons-ui from gridstudy and gridexplore.
Create declaration files during compilation for TypeScript files.
Make the tests works.
Add eslint command.

Notes : 
- we kept the old `eslint-config-react-app` from CRA in order to keep the same eslint rules, but the module is not maintained anymore and not compatible with some recent versions of TypeScript modules. That's why we installed specific older versions for these modules.
- when importing locally commons-ui in gridstudy / gridexplore, the types are not read very well after doing `npm prune --production`. I don't know why, and hope it works better when imported from npm, because the output files are exactly the same as MUI for example.
- `allowJS` seems to be broke
- compared to tsconfig of other projects, `esModuleInterop` is not set to `true`, because from what I understand it's not necessary for us, it produces less code and then `allowSyntheticDefaultImports` can remain `false` which is better to produce safe imports
- no TypeScript check while building / starting https://vitejs.dev/guide/features.html#typescript. Should do the same for eslint ?


package.json
```diff
@@ -8,17 +8,19 @@
...
+    "types": "./dist/index.d.ts",
     "files": [
         "dist"
     ],
     "scripts": {
         "start": "vite demo/ --config vite.config.mts",
-        "build": "npm install; vite build; npm prune --production",
-        "build:only": "vite build",
+        "build": "npm install && tsc --p ./tsconfig-build.json && vite build && npm prune --production",
+        "build:only": "tsc --p ./tsconfig-build.json && vite build",
```
TheMaskedTurtle, line 18: vite build cannot integrate ts compiling directly ?

package.json
```diff
@@ -82,21 +87,25 @@
         "react-router-dom": "^6.0.0",
         "ts-node": "^10.9.2",
         "type-fest": "^2.0.0",
-        "typescript": "^5.3.3",
+        "typescript": "5.1.6",
```
TheMaskedTurtle, line 90: Why ?

src/index.d.ts
```diff
@@ -0,0 +1,229 @@
...
+type Input = string | number;
+type Options = Array<{
+    id: string;
+    label: string;
+}>;
+
+interface AutocompleteInputProps
+    extends Omit<
+        AutocompleteProps<
+            string | { id: string; label: string },
```
TheMaskedTurtle, line 56: Maybe we can extract a type for this `string | { id: string; label: string }` ?

src/index.d.ts
```diff
@@ -0,0 +1,229 @@
...
+            boolean | undefined,
+            boolean | undefined,
+            boolean | undefined
+        >,
+        // we already defined them in our custom Autocomplete
+        'value' | 'onChange' | 'renderInput'
```
TheMaskedTurtle, line 62: We should be very careful with the Omit because it doesn't prevent to pass an object that contains these properties, and in all our component we put the remaining props at the end {...props} so if it is present it will override our values...

src/index.d.ts
```diff
@@ -0,0 +1,229 @@
...
+
+export const FieldErrorAlert: FunctionComponent;
+
+type TextFieldWithAdornmentProps = TextFieldProps & {
+    // variant already included in TextFieldProps
+    value: Input; // we override the default type of TextFieldProps which is unknown
+    adornmentPosition: string;
+    adornmentText: string;
+    handleClearValue?: () => void;
+};
```
TheMaskedTurtle, line 109: Maybe we should remain consistent between extended interfaces and types, just for readability

src/index.d.ts
```diff
@@ -0,0 +1,229 @@
...
+    name: string;
+    description: string;
+    type: string;
+    defaultValue: any;
+    possibleValues?: string[] | null;
+}
+
+interface FlatParametersProps extends Pick<TextFieldProps, 'variant'> {
+    paramsAsArray: Parameters[];
+    initValues: Record<string, any>;
```
TheMaskedTurtle, line 190: Record<string, unknown> ?

TheMaskedTurtle (commented):
A lot of types are just redefining react properties, not sure it is really valuable. Maybe we should just forward react types or even make our components generic ? 

Otherwise some warning points :
- We should be careful with the pattern Omit<>, it doesn't prevent the user of the component to pass objects containing these values, and in our component we propagate the remaining props at the end, it should be done at the beginning so that we're sure we override with our custom values.
- We should avoid any as much as possible, most of the time it could be replaced by unknown or Record<string, unknown>
- You can see in this [PR](https://github.com/gridsuite/gridapp-template/pull/57/files) that we identified other needs for types coming from commons-ui, especially types coming from oidc-client (js) with User and UserManager classes, and maybe some others like the definition of parameters coming from apps-metadata-server

flomillot:
> A lot of types are just redefining react properties, not sure it is really valuable. Maybe we should just forward react types or even make our components generic ?
> 
> Otherwise some warning points :
> 
> * We should be careful with the pattern Omit<>, it doesn't prevent the user of the component to pass objects containing these values, and in our component we propagate the remaining props at the end, it should be done at the beginning so that we're sure we override with our custom values.
> * We should avoid any as much as possible, most of the time it could be replaced by unknown or Record<string, unknown>
> * You can see in this [PR](https://github.com/gridsuite/gridapp-template/pull/57/files) that we identified other needs for types coming from commons-ui, especially types coming from oidc-client (js) with User and UserManager classes, and maybe some others like the definition of parameters coming from apps-metadata-server

As I said I just moved the file from gridstudy / gridexplore, the goal of the US was not to fix all the types of these files (some are bad I agree). The plan is to merge it quickly to allow other devs to use TypeScript in the project. I will do an other PR to clean this file with all your remarks.

src/index.d.ts
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+ * Section to export manual type declarations of .js and .jsx files
+ */
+
+export const TopBar: FunctionComponent<InferProps<typeof RawTopBar.propTypes>>;
+
+export function logout(dispatch: any, userManagerInstance: any): Promise<any | undefined>;
+
+export const DARK_THEME: String, LIGHT_THEME: String;
```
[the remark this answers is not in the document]
flomillot: Done in #357

src/index.d.ts
```diff
@@ -0,0 +1,232 @@
...
+
+interface SnackInputs {
+    messageTxt?: string;
+    messageId?: string;
+    messageValues?: { [key: string]: string };
```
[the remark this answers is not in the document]
flomillot: Will do in https://github.com/gridsuite/commons-ui/pull/357

src/index.d.ts
```diff
@@ -0,0 +1,227 @@
...
+export const elementType = {
+    DIRECTORY: 'DIRECTORY',
+    STUDY: 'STUDY',
+    FILTER: 'FILTER',
+    MODIFICATION: 'MODIFICATION',
+    CONTINGENCY_LIST: 'CONTINGENCY_LIST',
+    VOLTAGE_INIT_PARAMETERS: 'VOLTAGE_INIT_PARAMETERS',
+    SECURITY_ANALYSIS_PARAMETERS: 'SECURITY_ANALYSIS_PARAMETERS',
+    LOADFLOW_PARAMETERS: 'LOADFLOW_PARAMETERS',
+};
```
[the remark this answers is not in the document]
flomillot: It was the plan, will do in https://github.com/gridsuite/commons-ui/pull/357

src/index.d.ts
```diff
@@ -0,0 +1,227 @@
...
+    formProps?: Omit<RadioGroupProps, 'value' | 'onChange'>;
+}
+
+export const RadioInput: FunctionComponent<RadioInputProps>;
+
+interface SwitchInputProps {
+    name: string;
+    label?: string;
+    formProps?: Omit<SwitchInputProps, 'disabled'>;
+}
```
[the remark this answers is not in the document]
flomillot: Totally, will do in https://github.com/gridsuite/commons-ui/pull/357

src/index.d.ts
```diff
@@ -0,0 +1,227 @@
...
+    LOADFLOW_PARAMETERS: 'LOADFLOW_PARAMETERS',
+};
+
+export const OverflowableText: FunctionComponent<OverflowableTextProps>;
+
+interface CheckboxInputProps {
+    name: string;
+    label?: string;
+    formProps?: Omit<CheckboxInputProps, 'disabled'>;
+}
```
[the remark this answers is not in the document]
flomillot: Same

src/index.d.ts
```diff
@@ -0,0 +1,227 @@
...
+    label?: string;
+    formProps?: Omit<SwitchInputProps, 'disabled'>;
+}
+
+export const SwitchInput: FunctionComponent<SwitchInputProps>;
+
+export const SubmitButton: FunctionComponent<ButtonProps>;
+
+type CancelButtonProps = ButtonProps & {
+    color?: string;
```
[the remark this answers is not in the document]
flomillot: I guess one of the color wasn't in the MUI props ? x)

package.json
flomillot: Nope, I will change all these scripts in an other PR

package.json
flomillot: Answered in the PR description

src/index.d.ts
flomillot: Definitely, will do in #357

src/index.d.ts
flomillot: It does.

image

Or I missed something ? 

This screenshot if from gridstudy (main branch), but maybe it does not works correctly with the new branch. How did you get this result ?

src/index.d.ts
flomillot: What do you mean ?

src/index.d.ts
flomillot: Answered above for any / unknow

TheMaskedTurtle (approved):

src/index.d.ts
TheMaskedTurtle: yes if you do like that but if you do {...props} with an object containing value it does work , no ?

src/index.d.ts
```diff
@@ -0,0 +1,232 @@
...
+export { useTest } from './hooks/useTest';
+export { Test } from './utils/Test';
+
+/**
+ * Section to export manual type declarations of .js and .jsx files
+ */
+
+export const TopBar: FunctionComponent<InferProps<typeof RawTopBar.propTypes>>;
+
+export function logout(dispatch: any, userManagerInstance: any): Promise<any | undefined>;
```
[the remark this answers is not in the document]
TheMaskedTurtle: yes I'll be glad to discuss about that

src/index.d.ts
TheMaskedTurtle: I mean it's quite the same to do `interface TextFieldWithAdornmentProps extends TextFieldProps {` and except if in special cases there is a reason to use one or the other it's ok, but if there's no reason we can privilege one solution

src/index.d.ts
TheMaskedTurtle: Capture d’écran de 2024-02-23 13-02-55

flomillot: Ah yes weird...

flomillot: What is the correct solution then ?

src/index.d.ts
flomillot: Interface can't extend type, TS2312

src/index.d.ts
```diff
@@ -0,0 +1,232 @@
...
+export const elementType = {
+    DIRECTORY: 'DIRECTORY',
+    STUDY: 'STUDY',
+    FILTER: 'FILTER',
+    MODIFICATION: 'MODIFICATION',
+    CONTINGENCY_LIST: 'CONTINGENCY_LIST',
+    VOLTAGE_INIT_PARAMETERS: 'VOLTAGE_INIT_PARAMETERS',
+    SECURITY_ANALYSIS_PARAMETERS: 'SECURITY_ANALYSIS_PARAMETERS',
+    LOADFLOW_PARAMETERS: 'LOADFLOW_PARAMETERS',
+};
```
Tristan-WorkGH, line 222: use an `enum` maybe?

Tristan-WorkGH (changes requested):

package.json
```diff
@@ -8,17 +8,19 @@
...
         "start": "vite demo/ --config vite.config.mts",
-        "build": "npm install; vite build; npm prune --production",
+        "build": "npm install && tsc && vite build && npm prune --production",
         "build:only": "vite build",
         "prepublishOnly": "npm run build",
         "test": "jest",
         "test:coverage": "jest --coverage",
-        "test:watch": "jest --watch"
+        "test:watch": "jest --watch",
+        "lint": "eslint . --ext js,mjs,jsx,ts,mts,tsx --max-warnings 0"
```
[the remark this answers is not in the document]
flomillot: Nope because we want to check ./demo and some root files

src/index.d.ts
flomillot: Done in #357

Tristan-WorkGH (approved):

PR #362, opened 2024-02-29
Use proper icon for study element

dbraquart:
also added CASE element type.

Potential impacts of this MR in grid-explore:
- getElementIcon should/must be deleted and replaced by getFileIcon from commons-ui
- ElementType enum should/must be deleted and replaced by elementType from commons-ui (like it's done in grid-study)


FranckLecuyer (approved):
Code review: OK
Tests: OK

PR #371, opened 2024-03-11
rename elementType and remove duplicate

dbraquart:

PR #382, opened 2024-03-21
feat(equipment-type): make EquipmentType as enum usable in TS

TheMaskedTurtle:


sBouzols:
There is already a work about this in #371 no ?

TheMaskedTurtle:
> There is already a work about this in #371 no ?

@sBouzols  well spotted thank you ! I'm gonna synchronize with @flomillot !

src/utils/EquipmentType.ts
```diff
@@ -44,81 +44,117 @@ export const equipmentStyles = {
...
         tagLabel: 'equipment_search/busTag',
     },
-    SWITCH: {
-        name: 'SWITCH',
+    [EquipmentType.SWITCH]: {
+        name: EquipmentType.SWITCH,
         tagLabel: 'equipment_search/switchTag',
     },
+    [EquipmentType.VSC_CONVERTER_STATION]: undefined,
+    [EquipmentType.LCC_CONVERTER_STATION]: undefined,
```
flomillot, line 138: We can't define them also ?

src/utils/EquipmentType.ts
```diff
@@ -44,81 +44,117 @@ export const equipmentStyles = {
...
+}
+
 export const getEquipmentsInfosForSearchBar = (
-    equipmentsInfos,
-    getNameOrId
+    equipmentsInfos: Equipment[],
+    getNameOrId: (e: Identifiable) => string
 ) => {
-    return equipmentsInfos.flatMap((e) => {
+    return equipmentsInfos.flatMap((e): unknown => {
```
flomillot, line 155: Why `unknown` and not `Equipment` ?

src/utils/EquipmentType.ts
TheMaskedTurtle: I think they're both gathered under the HVDC_CONVERTER_STATION, but as I defined it as a Record<EquipmentType, > I had to define it with all the keys. I don't know where this object is used and not sure the tagLabel will make sense for them.

flomillot (approved):

PR #384, opened 2024-03-22
Fix DirectoryItemsInputProps interface

sBouzols:
use same type than the one define in sub component `DirectoryItemSelector`


TheMaskedTurtle (commented):
Why not to delete the properties that are just related to DirectoryItemsSelector and make DirectoryItemsInputProps extends DirectoryItempsSelectorProps ?

flomillot (approved):

PR #387, opened 2024-03-22
Add github action to check license header and dependencies license

sBouzols:
Run License header check action (check Mode)
Add dev dependency license-checker
Add script licenses-check to manage configuration from license-checker-config.json
Run this script command in CI.

We chose to not implement the License Header Fix mode for now.
~~If something need to be fix,~~
* ~~CI Fail~~
* ~~a PR is automatically created. User need to review and merge it~~
* ~~A comment is posted on the initial PR to point out this PR.~~

~~Notes : Manage JSX as well in config file because it's not supported by default.~~


.github/workflows/build.yml
```diff
@@ -2,6 +2,79 @@ name: CI
...
+                      commit_sha: context.sha,
+                      owner: context.repo.owner,
+                      repo: context.repo.repo,
+                    })
+                  ).data[0].number;
+                }
+              result-encoding: string
+
+          - name: Create Pull Request if necessary
+            if: github.ref != 'refs/heads/main' && ${{ steps.get_issue_number.outputs.result }} != ''
```
Tristan-WorkGH, line 42: create a new PR _if_ it's different of `main` branch? isn't the inverse?

sBouzols: nope, we don't want this step of this job to be executed at push event on main. This way it will create a PR only on push event on a branch with an existing PR.

.github/workflows/build.yml
```diff
@@ -2,6 +2,20 @@ name: CI
 on: [push]
 
 jobs:
+    licenses:
+        name: Check licenses headers and dependencies
```
antoinebhs, line 6: Check licenses headers
```suggestion
        name: Check licenses headers
```

.github/workflows/build.yml
```diff
@@ -2,6 +2,20 @@ name: CI
 on: [push]
 
 jobs:
+    licenses:
```
antoinebhs, line 5:
```suggestion
    license-headers:
```
?

antoinebhs (approved):

jamal-khey (approved):

PR #396, opened 2024-03-25
fix(directory-item-selector-props): remove unnecessary undefined type

TheMaskedTurtle:
Remove unnecessary undefined, as we always want to filter on equipment types.


dbraquart (approved):
code: ok
tests: not done

PR #400, opened 2024-03-26
Hides the application name on small screen

EstherDarkish:


src/components/TopBar/GridLogo.jsx
```diff
@@ -23,6 +23,9 @@ const styles = {
     clickable: {
         cursor: 'pointer',
     },
+    hideOnSmallScreen: {
+        display: { xs: 'none', lg: 'block' },
+    },
```
Tristan-WorkGH, line 28: place it in `title: {...}` style, the `style` property is passed to the `sx` inside `LogoText`

Tristan-WorkGH (changes requested):

src/components/TopBar/GridLogo.jsx
EstherDarkish: Thanks, that's better.

Tristan-WorkGH (approved):

PR #424, opened 2024-04-30
Change SVG mock for test

flomillot:
Better than the previous one, and remove its dependency.
Also fix a test, we were giving a raw component but the prop need to be a JSX to works.

### DIFF
--- a/src/components/TopBar/TopBar.test.js
+++ b/src/components/TopBar/TopBar.test.js
@@ -54,7 +54,7 @@ it('renders', () => {
                     <TopBar
                         appName="Demo"
                         appColor="#808080"
-                        appLogo={PowsyblLogo}
+                        appLogo={<PowsyblLogo/>}
                         onParametersClick={() => {}}
                         onLogoutClick={() => {}}
                         onLogoClick={() => {}}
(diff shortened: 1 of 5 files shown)


src/components/TopBar/TopBar.test.js
```diff
@@ -54,7 +54,7 @@ it('renders', () => {
                     <TopBar
                         appName="Demo"
                         appColor="#808080"
-                        appLogo={PowsyblLogo}
+                        appLogo={<PowsyblLogo />}
```
Tristan-WorkGH, line 57: But that give a rebuild of the logo at each evaluation no?

flomillot: I'm not sure, maybe, but that's how it works. 
https://github.com/gridsuite/commons-ui/blob/b49db7c8704b5cc18fe239bde6a5536b0d31aa1e/src/components/TopBar/GridLogo.jsx#L36

flomillot: https://github.com/gridsuite/gridstudy-app/blob/8aca6bf551b705fd4413ca525da77778abe74138/src/components/app-top-bar.jsx#L291-L297

jonenst (approved):

Tristan-WorkGH (approved):

PR #426, opened 2024-04-30
Allow Snackbar to be persistent

jamal-khey:
This pull request adds the ability to make Snackbar messages persistent. Previously, Snackbar messages would disappear after a certain amount of time, but now they can be set to stay on the screen until dismissed by the user.


src/hooks/useSnackMessage.ts
```diff
@@ -17,65 +22,61 @@ interface SnackInputs {
     headerTxt?: string;
     headerId?: string;
     headerValues?: Record<string, string>;
+    persistent?: boolean;
```
TheMaskedTurtle, line 25: Why don't we just extend notistack type ? Then it would be persist, not persistent. But I think we shouldn't be intrusive into the notistack API, we just want to extend it a bit

src/hooks/useSnackMessage.ts
```diff
@@ -95,7 +96,7 @@ function makeSnackbar(
             header,
             enqueueSnackbar,
             level,
-            persistent
+            snackInputs.persistent ? snackInputs.persistent : false
```
TheMaskedTurtle, line 99: If you want to eliminate nullish values just do `snackInputs.persistent ?? false` but it's not really necessary as for notistack default value is false already

TheMaskedTurtle (changes requested):
We're trying to just extend a bit the notistack API here with our snackInfo, snackWarn, etc.. methods, so it would be nice to not limit it. If so, it wouldn't require any dev to add the persist behavior as it is already in the notistack API.

src/hooks/useSnackMessage.ts
jamal-khey: i exetended OptionsObject interface that define the attributes we want

src/hooks/useSnackMessage.ts
jamal-khey: i reused the OptionsObject type here

src/hooks/useSnackMessage.ts
```diff
@@ -89,14 +91,12 @@ function makeSnackbar(
...
-            message,
-            header,
-            enqueueSnackbar,
-            level,
-            persistent
-        );
+        displayMessageWithSnackbar(message, header, enqueueSnackbar, level, {
+            key: snackInputs.key,
+            persist: snackInputs.persist,
+        });
```
TheMaskedTurtle, line 99: You could pass snackInputs directly, to make sure that if we want additional properties we can still pass them

src/hooks/useSnackMessage.ts
```diff
@@ -143,9 +143,9 @@ function displayMessageWithSnackbar(
         }
         fullMessage += message;
     }
-    enqueueSnackbar(fullMessage, {
+    return enqueueSnackbar(fullMessage, {
         variant: level,
-        persist: persistent,
+        ...enqueOption,
```
TheMaskedTurtle, line 148:
```suggestion
        ...enqueOption,
        variant,
        style: { whiteSpace: 'pre-line' },
```
To make sure you override variant

src/hooks/useSnackMessage.ts
```diff
@@ -131,7 +131,7 @@ function displayMessageWithSnackbar(
     header: string,
     enqueueSnackbar: EnqueueSnackbar,
     level: BaseVariant,
```
TheMaskedTurtle, line 133: You can rename variant

src/hooks/useSnackMessage.ts
```diff
@@ -131,7 +131,7 @@ function displayMessageWithSnackbar(
...
-    persistent: boolean
+    enqueOption: OptionsObject
```
TheMaskedTurtle, line 134: typo: enqueueOptions

TheMaskedTurtle (commented):
It's not really on your dev but the makeSnackbar method kind of mixes the logic of building the message and the one that just enqueue the message, maybe you could split them to make it more readable.
Something like that for the enqueue logic 
```
const enqueue = useCallback(
        (variant: BaseVariant, snackInputs: SnackInputs) => {
            const message = makeMessage(intlRef, snackInputs);
            if (message !== undefined) {
                enqueueSnackbar(message, {
                    ...snackInputs,
                    variant,
                    style: { whiteSpace: 'pre-line' },
                });
            }
        },
        [enqueueSnackbar, intlRef]
    );
```
And then you keep the makeMessage with the current message logic.

src/hooks/useSnackMessage.ts
jamal-khey: done

src/hooks/useSnackMessage.ts
jamal-khey: done

src/hooks/useSnackMessage.ts
jamal-khey: done

src/hooks/useSnackMessage.ts
jamal-khey: snackInputs contains others properties that are not needed for enqueueSnackbar, that why i extracted the required one.
but i suppose enqueueSnackbar don't care if we give it more than what it needs

src/hooks/useSnackMessage.ts
```diff
@@ -23,81 +28,57 @@ export interface UseSnackMessageReturn {
     snackError: (snackInputs: SnackInputs) => void;
     snackWarning: (snackInputs: SnackInputs) => void;
     snackInfo: (snackInputs: SnackInputs) => void;
+    closeSnackbar: (key?: SnackbarKey) => void;
```
TheMaskedTurtle, line 31: Maybe `typeof closeSnackbar` as we just forward notistack method

src/hooks/useSnackMessage.ts
```diff
@@ -23,81 +28,57 @@ export interface UseSnackMessageReturn {
...
-        snackInputs.messageId,
-        snackInputs.messageValues
-    );
-    const header = checkAndTranslateIfNecessary(
-        intlRef,
-        snackInputs.headerTxt,
-        snackInputs.headerId,
-        snackInputs.headerValues
-    );
-    if (message !== null && header !== null) {
```
TheMaskedTurtle, line 92: I'm afraid we have a regression here, because previously if one is null we don't display the snackbar. We should keep it to be sure we have no important regression. Even if it is certainly not the main use case.

src/hooks/useSnackMessage.ts
```diff
@@ -126,13 +107,22 @@ function checkInputs(txt?: string, id?: string, values?: any) {
...
 
-function displayMessageWithSnackbar(
-    message: string,
-    header: string,
-    enqueueSnackbar: EnqueueSnackbar,
-    level: BaseVariant,
-    persistent: boolean
-) {
+function makeMessage(
+    intlRef: React.MutableRefObject<IntlShape>,
```
TheMaskedTurtle, line 111: Might be better to remove the React. to avoid importing the whole React module

TheMaskedTurtle (changes requested):

src/hooks/useSnackMessage.ts
```diff
@@ -6,11 +6,16 @@
...
+    BaseVariant,
+    OptionsObject,
+    SnackbarKey,
+    useSnackbar,
+} from 'notistack';
 import { useIntlRef } from './useIntlRef';
 import { IntlShape } from 'react-intl';
 
-interface SnackInputs {
+interface SnackInputs extends OptionsObject {
```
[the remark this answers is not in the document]
jamal-khey: done

src/hooks/useSnackMessage.ts
jamal-khey: done

src/hooks/useSnackMessage.ts
jamal-khey: done

demo/src/app.jsx
```diff
@@ -245,6 +245,41 @@ function SnackInfoButton() {
...
 
+function PermanentSnackButton() {
+    const { snackInfo, closeSnackbar } = useSnackMessage();
+    const [snackKey, setSnackKey] = useState(undefined);
+    return (
+        <>
+            <Button
+                variant="contained"
+                color="info"
+                style={{ float: 'left', margin: '5px' }}
```
souissimai, line 256: make style button sx={style.button}

TheMaskedTurtle (approved):

demo/src/app.jsx
jamal-khey: done

souissimai (approved):

PR #428, opened 2024-05-02
Add missing load type field and change order of fields for two windings tranformer in expert filter

FranckLecuyer:


dbraquart (approved):
code: ok
tests: ok
 - 2WT order has changed
 - Load: type added and filtering ok

PR #430, opened 2024-05-03
Make element-search-dialog easier to customize

klesaulnier:


src/components/ElementSearchDialog/element-search-dialog.tsx
```diff
@@ -0,0 +1,170 @@
...
+                            typeof option === 'string' ||
+                            typeof value === 'string'
+                        ) {
+                            return option === value;
+                        } else {
+                            return option.id === value.id;
+                        }
+                    }}
+                    options={isLoading ? [] : elementsFound}
+                    loading={isLoading}
```
sBouzols, line 120: If you add `loading` please add `loadingText` as well and some keys for translation purpose

src/components/ElementSearchDialog/element-search-dialog.tsx
```diff
@@ -0,0 +1,170 @@
...
+        onClose,
+        searchingLabel,
+        searchTerm,
+        onSearchTermChange,
+        onSelectionChange,
+        elementsFound, // [{ label: aLabel, id: anId }, ...]
+        renderElement,
+        searchTermDisabled,
+        searchTermDisableReason,
+        isLoading,
```
sBouzols, line 50: Why do we need to expose this prop to the owner of `ElementSearchDialog` ?
Maybe we could set a local state to true when `onSearchTermChange` is called then set it to false when elementsFound changes ?
I think It's complex if we let the component owner manage this state no ?
At least name is `loading` as the Autocomplete prop ?

sBouzols:
When testing I saw that we show the "loading..." text when input field is empty, that doesn't happen at the moment in gridstudy

src/hooks/useDebounce.ts
```diff
@@ -8,7 +8,10 @@
 import { useEffect, useMemo } from 'react';
 import { debounce } from '@mui/material';
 
-export const useDebounce = (func: (...args: any[]) => any, delay = 700) => {
+export const useDebounce = <T extends (...args: any[]) => any>(
```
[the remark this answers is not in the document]
klesaulnier: yes, the return type was not correct before

before, the returned method had this type :
**((...args: any[]) => any) & Cancelable**

now, the returned method has this type : 
**T & Cancelable**


meaning if I pass a "(param: string) => void" method as an argument, I had this as a returned type : 
**((...args: any[]) => any) & Cancelable**

now I have this : 
**((value: string) => void) & Cancelable**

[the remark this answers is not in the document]
klesaulnier: Also, this is the type used by material-ui "debounce"

src/components/ElementSearchDialog/element-search-dialog.tsx
klesaulnier: I'm not sure I understand, you don't want 
` noOptionsText={intl.formatMessage({
       id: 'element_search/noResult',
  })}`
to be in commons-ui anymore ?

src/components/ElementSearchDialog/element-search-dialog.tsx
klesaulnier: I think it's way better to handle "isLoading" when we make the fetch, otherwise we will have some troubles handling errors or edge cases

for "isLoading" naming, I see your point, but I think it's more clear to call it "isLoading"
Can we ask a 3rd person to decide with us ?

klesaulnier:
> When testing I saw that we show the "loading..." text when input field is empty (enter a character, then remove it), that doesn't happen at the moment in gridstudy

Before, emptying the field would make the suggestions empty, now we are displaying the history
But you're right, maybe we should display them without having to way for the debounce to actually execute if the use input is empty

src/hooks/useDebounce.ts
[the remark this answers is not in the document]
sBouzols: thanks a lot for this explanation !

src/components/ElementSearchDialog/element-search-dialog.tsx
sBouzols: I'm talking about `loadingText` (which is by default "Loading..." ) not `noOptionsText`.
It will not be translated in french by MUI, I think we must manage this prop with `loading` prop

sBouzols:
> > When testing I saw that we show the "loading..." text when input field is empty (enter a character, then remove it), that doesn't happen at the moment in gridstudy
> 
> Before, emptying the field would make the suggestions empty, now we are displaying the history But you're right, maybe we should display them without having to way for the debounce to actually execute if the use input is empty

Yex I think it could be a problem if we fetch something with an empty string you know.
I think we want to avoid this for any `onSearchTermChange` callback prop then maybe just don't call it in this component if searchTerm is empty ?
image

src/components/ElementSearchDialog/element-search-dialog.tsx
```diff
@@ -0,0 +1,170 @@
...
+    } = props;
+
+    const displayedValue = useMemo(() => {
+        return searchTermDisabled || searchTermDisableReason
+            ? searchTermDisableReason
+            : searchTerm;
```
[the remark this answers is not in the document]
klesaulnier: Hmm I'm not sure here, he we could have a behaviour where searchTermDisabled is "true" and searchTermDisableReason is empty

It means we would disable the component, but still display "searchTerm" in the textInput

This was the previous behaviour if I'm not wrong

src/components/ElementSearchDialog/element-search-dialog.tsx
souissimai: ?on which case we can have troubles or errors when using state instead of props

PR #436, opened 2024-05-06
Add filter API

flomillot:
Migrate the filter API functions from gridexplore to commons-ui.
Add basic API util functions to be able to fetch servers from commons-ui.
Export a function to set the common store to give access to the user token.


src/services/explore.ts
```diff
@@ -0,0 +1,55 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { UUID } from 'crypto';
+import { backendFetch } from './utils';
+
+const PREFIX_EXPLORE_SERVER_QUERIES =
+    import.meta.env.VITE_API_GATEWAY + '/explore';
+
+export function createFilter(
```
TheMaskedTurtle, line 14: Why don't you define these methods as hooks ? So you can import the token directly here and you make it transparent for the components that use APIs ?

package.json
```diff
@@ -59,7 +59,9 @@
         "react-hook-form": "^7.51.2",
         "react-intl": "^6.6.4",
         "react-papaparse": "^4.1.0",
+        "react-redux": "^9.1.2",
```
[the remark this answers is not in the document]
flomillot: I know, both in peer deps and dev deps

src/services/explore.ts
flomillot: Personnaly I prefer like that, otherwise that means we encapsulate every fetch in a hook, it's a bit of boilerplate code for me.
We can discuss about to take a decision if you want.

package.json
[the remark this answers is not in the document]
TheMaskedTurtle: Ah yes sorry I thought that was in deps and dev-deps

src/services/explore.ts
TheMaskedTurtle: I agree it's a bit of additionnal code, but I don't like that you need to retrieve userToken in functional components. Functional components should remain functional and when you call createFilter you want to pass only functional arguments to this method, the management of the token should be handled somewhere else in a purely technical layer.

flomillot: Ok let's discuss about it

src/services/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (!(typeof init == 'undefined' || typeof init == 'object')) {
+        throw new TypeError(
+            'Argument 2 of backendFetch is not an object : ' + typeof init
+        );
+    }
+    const initCopy = Object.assign({}, init);
+    initCopy.headers = new Headers(initCopy.headers || {});
+    initCopy.headers.append(
+        'Authorization',
+        'Bearer ' + token ?? getUserToken()
```
TheMaskedTurtle, line 25: Why `token ?? getUserToken()` ? I think we just want to use getUserToken and remove all the tokens in argument no ?

src/components/filter/utils/filter-api.ts
```diff
@@ -21,14 +22,8 @@ export const saveExplicitNamingFilter = (
...
     handleClose: () => void,
-    createFilter: (
-        filter: any,
-        name: string,
-        description: string,
-        activeDirectory: any
-    ) => Promise<void>,
-    saveFilter: (filter: any, name: string) => Promise<void>,
-    activeDirectory?: UUID
+    activeDirectory: UUID | undefined,
```
Tristan-WorkGH, line 25:
```suggestion
    activeDirectory?: UUID,
```

src/services/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import { getUserToken } from '../redux/commonStore.ts';
+
+export const backendFetch = (url: string, init: any, token?: string) => {
+    const initCopy = prepareRequest(init, token);
+    return safeFetch(url, initCopy);
+};
+
+const prepareRequest = (init: any, token?: string) => {
+    if (!(typeof init == 'undefined' || typeof init == 'object')) {
```
Tristan-WorkGH, line 16:
```suggestion
    if (!(typeof init === 'undefined' || typeof init === 'object')) {
```

src/services/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+        throw new TypeError(
+            'Argument 2 of backendFetch is not an object : ' + typeof init
```
Tristan-WorkGH, line 18: `init` argument is the first one
```suggestion
            'Argument 1 of backendFetch is not an object: ' + typeof init
```

src/services/explore.ts
```diff
@@ -0,0 +1,55 @@
...
+    newFilter: any,
+    name: string,
+    description: string,
+    parentDirectoryUuid: UUID | undefined,
```
Tristan-WorkGH, line 18:
```suggestion
    parentDirectoryUuid?: UUID,
```

src/services/utils.ts
flomillot: As it's supposed to be a temporary solution, I kept my work with the token in functions args and made it optional, in case we need it later.
But I can revert all this.

flomillot: It's also how it's done in gridexplore / gridstudy.
https://github.com/gridsuite/gridexplore-app/blob/a01ac4c3d1804830a062e7d897be6dc39b60d0b2/src/utils/rest-api.js#L112

Tristan-WorkGH (changes requested):

src/components/filter/utils/filter-api.ts
flomillot: Yah I think I will revert this

jonenst (approved):

src/services/utils.ts
flomillot: I kept the same as in gridstudy / explore but yeah I correct it

src/services/utils.ts
flomillot: actually it's the second arg of backendFetch :sweat_smile:

flomillot: maybe move it in backendFetch instead ?

src/services/explore.ts
flomillot: Decision has been made to use a global var to store the redux store and have access to the token from the API files directly.
Temporary solution before refactoring the whole token management.

src/services/utils.ts
flomillot: i just change the message

src/services/utils.ts
TheMaskedTurtle: yeah it's a bit weird to expose this argument in all the APIs but then no components should use it because it's done in the 'authentication layer'. I would rather remove it, but as it was in the apps already, it's up to you.

flomillot: Breifly discussed with @jonenst , I think we will keep it

src/services/utils.ts
```diff
@@ -20,10 +20,8 @@ const prepareRequest = (init: any, token?: string) => {
     }
     const initCopy = Object.assign({}, init);
     initCopy.headers = new Headers(initCopy.headers || {});
-    initCopy.headers.append(
-        'Authorization',
-        'Bearer ' + token ?? getUserToken()
-    );
+    const tokenCopy = token ? token : getUserToken();
```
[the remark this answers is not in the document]
thangqp: If we consider empty string, someone can pass also blank string.. so ?? or || can not resolve the problem.. i think test with undefined as the actual code is enough..

thangqp (approved):
Code OK, test OK with https://github.com/gridsuite/gridexplore-app/pull/424

src/services/utils.ts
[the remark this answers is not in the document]
flomillot: const tokenCopy = token ? token : getUserToken();
and
const tokenCopy = token ?? getUserToken();
arent the same ?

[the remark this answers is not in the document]
thangqp: ah YES! the both is falsy check..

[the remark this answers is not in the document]
TheMaskedTurtle: yes ?? eliminate only nullish (null and undefined) values, whereas the other code eliminates falsy values. As this part of the code is not used it's hard to know what we should check or not ^^ But yes that was just to raise awarness that both semantics are not equivalent and usually I prefer ?? for most cases, testing falsiness can be ambiguous and lead to bugs in js.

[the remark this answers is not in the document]
flomillot: Ok I didn't know (or forgot :) )
All of these utility functions comes from gridstudy / gridexplore as they are.
But I agree I prefer ?? here, I change it.

src/services/utils.ts
Tristan-WorkGH: > actually it's the second arg of backendFetch 😅

ahhh okay !  
yes effectively

> maybe move it in backendFetch instead ?

maybe 🤔

> i just change the message

well oaky

src/services/utils.ts
Tristan-WorkGH: gridstudy code isn't all good 🤣

Tristan-WorkGH (approved):

PR #438, opened 2024-05-07
Restore convert filters to explicit naming feature

Meklo:


src/components/filter/explicit-naming/explicit-naming-filter-form.tsx
```diff
@@ -186,19 +212,50 @@ const ExplicitNamingFilterForm: FunctionComponent = () => {
...
+    const onStudySelected = (studyUuid: UUID) => {
+        exportFilter(studyUuid, sourceFilterForExplicitNamingConversion?.id)
+            .then((matchingEquipments: any) => {
+                setValue(
+                    FILTER_EQUIPMENTS_ATTRIBUTES,
+                    matchingEquipments.map((equipment: any) => ({
+                        [FieldConstants.AG_GRID_ROW_UUID]: uuid4(),
+                        [FieldConstants.EQUIPMENT_ID]: equipment.id,
+                        [DISTRIBUTION_KEY]: equipment.distributionKey,
+                    }))
```
dbraquart, line 224:
```suggestion
                    matchingEquipments.length === 0
                        ? makeDefaultTableRows()
                        : matchingEquipments.map((equipment: any) => ({
                              [FieldConstants.AG_GRID_ROW_UUID]: uuid4(),
                              [FieldConstants.EQUIPMENT_ID]: equipment.id,
                              [DISTRIBUTION_KEY]: equipment.distributionKey,
                          }))
```

dbraquart (commented):
code: ok, the report of "Convert" code is OK
tests: works fine, but I found an issue (not a regression)
Test case: 
- convert a filter with a study where we have some matching elements
- choose another study where there is no match : the previous list is still displayed.
This is very confusing. When we set an empty array, the table is not updated. I made a suggestion to display the default table... don't know if we can do better. It's up to you if you want to fix it, or leave it for later (in this case I will approve the PR as is).

src/components/filter/explicit-naming/explicit-naming-filter-form.tsx
Meklo: Done

dbraquart (approved):
code: ok
tests: not done on last commit

PR #455, opened 2024-05-23
Add mutualized api functions

Meklo:
Removes the FilterContext, replaced by the following functions for API calls related to fetching directory data : 
- fetchRootFolders
- fetchDirectoryContent
- fetchDirectoryElementPath
- elementExists
- fetchElementsInfos

The following method used to retrieve the application metadata :   
- fetchAppsAndUrls (renamed to fetchAppsMetadata)

Also:
- downloadZipFile (renamed to downloadFile)

Also adds common functions needed for data fetching mechanisms
- backendFetch
- backendFetchJson
- backendFetchFile
- backendFetchText


src/services/explore.ts
```diff
@@ -53,3 +57,40 @@ export function saveFilter(filter: any, name: string, token?: string) {
...
+
+    const params = [
+        ...idsParams,
+        ...equipmentTypesParams,
+        ...elementTypesParams,
+    ];
+
+    const urlSearchParams = new URLSearchParams(params);
+
+    const url = `${PREFIX_EXPLORE_SERVER_QUERIES}/v1/explore/elements/metadata?${urlSearchParams}`;
```
Tristan-WorkGH, line 90: `getRequestParamFromList()` return a `URLSearchParams`...
```suggestion
    const urlSearchParams = getRequestParamFromList([
        [ids.filter((id) => id), 'ids'], // filter falsy elements
        [equipmentTypes, 'equipmentTypes'],
        [elementTypes, 'elementTypes'],
    ]/*.filter(tuple => tuple[0])*/);

    const url = `${PREFIX_EXPLORE_SERVER_QUERIES}/v1/explore/elements/metadata?${urlSearchParams}`;
```

Or else `getRequestParamFromList(...): [string, string][]`, then:
```typescript
    const idsParams = getRequestParamFromList(
        ids.filter((id) => id), // filter falsy elements
        'ids'
    );

    const equipmentTypesParams = equipmentTypes
        ? getRequestParamFromList(equipmentTypes, 'equipmentTypes')
        : [];

    const elementTypesParams = elementTypes
        ? getRequestParamFromList(elementTypes, 'elementTypes')
        : [];

    const params = new URLSearchParams([
        ...idsParams,
        ...equipmentTypesParams,
        ...elementTypesParams,
    ]).toString();

    const url = `${PREFIX_EXPLORE_SERVER_QUERIES}/v1/explore/elements/metadata?${params}`;
```

src/services/directory.ts
flomillot, line 1: As discussed, should compare between gridstudy and gridexplore to take the best / newer code and eventually refactor it

flomillot (changes requested):

src/services/directory.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export function fetchDirectoryContent(directoryUuid: UUID, types?: string[]) {
+    console.info("Fetching Folder content '%s'", directoryUuid);
+
+    // Add params to Url
+    const typesParams = types
+        ? getRequestParamFromList(types, 'elementTypes')
+        : [];
+    const urlSearchParams = new URLSearchParams(typesParams);
```
Tristan-WorkGH, line 45: `getRequestParamFromList()` already return `URLSearchParams`

Tristan-WorkGH (changes requested):
part 1

src/services/apps-metadata.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Copyright (c) 2024, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+export function fetchEnv() {
+    return fetch('env.json').then((res) => res.json());
+}
```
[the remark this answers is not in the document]
Tristan-WorkGH: ping @jonenst @flomillot

what do you think?

src/index.ts
```diff
@@ -216,4 +216,23 @@ export {
...
+    downloadZipFile,
+} from './services/utils';
+export { fetchAppsAndUrls, fetchEnv } from './services/apps-metadata';
+export { fetchElementsInfos } from './services/explore';
+export {
+    fetchRootFolders,
+    fetchDirectoryContent,
+    fetchDirectoryElementPath,
+    elementExists,
+} from './services/directory';
```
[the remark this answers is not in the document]
flomillot: But i'm pretty sure people will continue to add some in the root index...

src/services/utils.ts
```diff
@@ -80,3 +114,33 @@ const parseError = (text: string) => {
         return null;
     }
 };
+
+export const getRequestParamFromList = (
+    params: string[],
+    paramName: string
+) => {
+    return new URLSearchParams(
+        params?.length ? params.map((param) => [paramName, param]) : []
```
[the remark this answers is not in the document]
flomillot: (agree for the length)

src/services/apps-metadata.ts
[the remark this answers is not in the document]
flomillot: IMO the function should stay here, we can eventually put the location in the common redux store

src/services/apps-metadata.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function fetchAppsAndUrls() {
```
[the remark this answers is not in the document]
Tristan-WorkGH: okay to rename, but the name would be then `(get|fetch)AppsMetadataServerUrl`.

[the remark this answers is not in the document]
flomillot: Nope we keep the same mistake, check what the fetch do !

src/services/utils.ts
```diff
@@ -12,13 +12,47 @@ export const backendFetch = (url: string, init: any, token?: string) => {
...
+    return backendFetch(url, init, token).then((safeResponse) =>
+        safeResponse.text()
+    );
+};
+
+export const backendFetchFile = (url: string, init?: any, token?: string) => {
+    return backendFetch(url, init, token).then((safeResponse) =>
+        safeResponse.blob()
+    );
+};
```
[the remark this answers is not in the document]
Meklo: `backendFetchText`, `backendFetchFile` `backendFetch` were accepting all 200 range return codes, won't restricting it to only 204 be prone to issues ?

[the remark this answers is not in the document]
flomillot: I would let as it is personnaly

src/services/apps-metadata.ts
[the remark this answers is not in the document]
Tristan-WorkGH: huh... I don't think that is what we want to do here: if i don't mistake, `env.json` is meant to be possibly dynamic if we want without having to reload/rebuild the app/container.

src/services/utils.ts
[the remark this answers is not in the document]
Tristan-WorkGH: That true, then set function parameters with `?`.

src/services/utils.ts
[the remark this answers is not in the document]
Tristan-WorkGH: Normally no, because a response 204 specifically say that there is no content in the response.  
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses  
> [204 No Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204)
    There is no content to send for this request, but the headers may be useful. The user agent may update its cached headers for this resource with the new ones.

src/hooks/predefined-properties-hook.ts
```diff
@@ -61,10 +55,9 @@ export const usePredefinedProperties = (
     const [equipmentPredefinedProps, setEquipmentPredefinedProps] =
         useState<PredefinedProperties>({});
     const { snackError } = useSnackMessage();
-    const { fetchAppsAndUrls } = useContext(FilterContext);
 
     useEffect(() => {
-        if (fetchAppsAndUrls && type !== null) {
+        if (type !== null) {
             fetchPredefinedProperties(type, fetchAppsAndUrls)
```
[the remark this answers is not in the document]
Meklo: Done

src/services/apps-metadata.ts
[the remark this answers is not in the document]
Meklo: Yea I see, it's about better reflecting the finality of the function instead of indicating the resource from which we get it from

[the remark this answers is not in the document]
Tristan-WorkGH: hum, yes fetchAppsMetadata
> .then((env) => fetch(env.appsMetadataServerUrl + '/apps-metadata.json'))

src/services/utils.ts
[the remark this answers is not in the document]
Meklo: Alternatively we could set a default value to params as follow :
`export const getRequestParamFromList = (
    params: string[] = [],
    paramName: string
) `

src/services/directory.ts
Meklo: Done with the input of @Tristan-WorkGH

src/index.ts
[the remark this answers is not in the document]
Meklo: Done

src/services/utils.ts
```diff
@@ -12,13 +12,47 @@ export const backendFetch = (url: string, init: any, token?: string) => {
...
+
+const FILE_TYPE = {
+    ZIP: 'ZIP',
+};
```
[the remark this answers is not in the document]
Meklo: Done

src/services/utils.ts
```diff
@@ -12,13 +12,47 @@ export const backendFetch = (url: string, init: any, token?: string) => {
...
+export const downloadZipFile = (blob: Blob, fileName: string) => {
+    downloadFile(blob, fileName, FILE_TYPE.ZIP);
+};
+
+const downloadFile = (blob: Blob, filename: string, type: string) => {
```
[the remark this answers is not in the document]
Meklo: Done

src/services/utils.ts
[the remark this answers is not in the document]
Meklo: I just noticed gridexplore's version didn't have the condition on 204, so I guess it's alright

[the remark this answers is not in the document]
Meklo: Done

src/services/explore.ts
Meklo: I went for the second option while removing the ternary operator on calls to `getRequestParamFromList`

src/services/directory.ts
Meklo: Cleaned it

src/services/apps-metadata.ts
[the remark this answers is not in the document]
Tristan-WorkGH: I see with Jon, and `env.json` like others files in `public` folders are meant to be accessible from "http://host/env.json", while the content must be fetched each time because the content can be modified by the deployment.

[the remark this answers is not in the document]
flomillot: Being dynamic does not mean the location change. And it's the same problem being here or in gridstudy no ?
Or I missed something ? I admit I don't understand the problem in this case.

src/services/metadata.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface Metadata {
+    name: string;
+    url: string;
+    appColor: string;
+    hiddenInAppsMenu: boolean;
+    resources: unknown;
+}
+
+export interface StudyMetadata extends Metadata {
```
flomillot, line 18: Now we have StudyMetadata and MetadataStudy.
Maybe merge apps-metadata.ts and metadata.ts ?

src/services/directory.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export function fetchDirectoryContent(directoryUuid: UUID, types?: string[]) {
+    console.info("Fetching Folder content '%s'", directoryUuid);
+
+    // Add params to Url
+    const urlSearchParams = getRequestParamFromList(
+        types,
+        'elementTypes'
+    ).toString();
+    const fetchDirectoryContentUrl = `${PREFIX_DIRECTORY_SERVER_QUERIES}/v1/directories/${directoryUuid}/elements?${urlSearchParams}`;
```
flomillot, line 47: `types` is optional

GridExplore : https://github.com/gridsuite/gridexplore-app/blob/4c6276610be67ccee01708cfecc9d2f63ddc28fc/src/utils/rest-api.js#L263-L265

An URL should not finish with `?` if no parameters, no ?

src/services/directory.ts
```diff
@@ -0,0 +1,77 @@
...
+    backendFetch,
+    backendFetchJson,
+    getRequestParamFromList,
+} from './utils';
+import { UUID } from 'crypto';
+
+const PREFIX_DIRECTORY_SERVER_QUERIES =
+    import.meta.env.VITE_API_GATEWAY + '/directory';
+
+function getPathUrl(elementUuid: UUID) {
```
flomillot, line 18: Do we really need this ? IMO it lower the readibility. The function is one line and called only once...

src/services/directory.ts
flomillot: See my remarks on fetchDirectoryContent

src/services/directory.ts
Tristan-WorkGH: if we want to be clean, yes, else it have no impact as http servers today will parse an empty map of parameters

src/services/utils.ts
Tristan-WorkGH, line 1: Look for
  * [`prepareRequest`](https://github.com/gridsuite/gridadmin-app/blob/664aee0be5529417fa7a1920fbcbeb750cd0fc5e/src/utils/api-rest.ts#L46-L57)
  * [`handleError`](https://github.com/gridsuite/gridadmin-app/blob/664aee0be5529417fa7a1920fbcbeb750cd0fc5e/src/utils/api-rest.ts#L26-L44)

src/services/utils.ts
```diff
@@ -80,3 +121,22 @@ const parseError = (text: string) => {
...
+    document.baseURI
+        .replace(/^http:\/\//, 'ws://')
+        .replace(/^https:\/\//, 'wss://');
+
+export const FetchStatus = {
+    SUCCEED: 'SUCCEED',
+    FAILED: 'FAILED',
+    IDLE: 'IDLE',
+    RUNNING: 'RUNNING',
+};
```
Tristan-WorkGH, line 142: It has nothing to do here, as it's an enum for components, not for api.

src/services/utils.ts
```diff
@@ -12,11 +12,52 @@ export const backendFetch = (url: string, init: any, token?: string) => {
...
+    return (await backendFetch(url, init, token)).blob();
+}
+
+export enum FileType {
+    ZIP,
+}
+
+export const downloadZipFile = (blob: Blob, fileName: string) => {
+    downloadFile(blob, fileName, FileType.ZIP);
+};
```
Tristan-WorkGH, line 45: obsolete, to delete

src/services/metadata.ts
```diff
@@ -0,0 +1,39 @@
...
+        [networkElementType: string]: PredefinedProperties;
+    };
+    defaultCountry?: string;
+}
+
+const isStudyMetadata = (metadata: MetadataJson): metadata is StudyMetadata => {
+    return metadata.name === 'Study';
+};
+
+export function fetchStudyMetadata() {
```
Tristan-WorkGH, line 30:
```suggestion
export function fetchStudyMetadata(): Promise<MetadataStudy> {
```

src/services/metadata.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface StudyMetadata extends Metadata {
+    name: 'Study';
+    predefinedEquipmentProperties: {
+        [networkElementType: string]: PredefinedProperties;
+    };
+    defaultCountry?: string;
+}
+
+const isStudyMetadata = (metadata: MetadataJson): metadata is StudyMetadata => {
```
Tristan-WorkGH, line 26:
```suggestion
const isStudyMetadata = (metadata: MetadataJson): metadata is MetadataStudy => {
```

src/services/explore.ts
```diff
@@ -53,3 +57,40 @@ export function saveFilter(filter: any, name: string, token?: string) {
         token
     );
 }
+
+export function fetchElementsInfos(
+    ids: UUID[],
+    elementTypes?: string[],
+    equipmentTypes?: string[]
+) {
```
Tristan-WorkGH, line 65:
```suggestion
): Promise<???> {
```
we don't want a `Promise<any>`.

src/services/utils.ts
```diff
@@ -12,11 +12,52 @@ export const backendFetch = (url: string, init: any, token?: string) => {
...
 
-export const backendFetchJson = (url: string, init: any, token?: string) => {
-    const initCopy = prepareRequest(init, token);
-    return safeFetch(url, initCopy).then((safeResponse) =>
-        safeResponse.status === 204 ? null : safeResponse.json()
+export async function backendFetchJson(
+    url: string,
+    init?: any,
+    token?: string
+) {
```
Tristan-WorkGH, line 19: @TheMaskedTurtle what do you think that we do
```suggestion
): Promise<unknown> {
```
to force a check or explicit cast where it's use?

src/services/directory.ts
```diff
@@ -0,0 +1,77 @@
...
+        'elementTypes'
+    ).toString();
+    const fetchDirectoryContentUrl = `${PREFIX_DIRECTORY_SERVER_QUERIES}/v1/directories/${directoryUuid}/elements?${urlSearchParams}`;
+    return backendFetchJson(fetchDirectoryContentUrl, {
+        method: 'get',
+        headers: { 'Content-Type': 'application/json' },
+    });
+}
+
+export function fetchDirectoryElementPath(elementUuid: UUID) {
```
Tristan-WorkGH, line 54:
```suggestion
export function fetchDirectoryElementPath(elementUuid: UUID): Promise<???> {
```
we don't want a `Promise<any>`.

src/services/directory.ts
```diff
@@ -0,0 +1,77 @@
...
+        'elementTypes'
+    ).toString();
+    const fetchRootFoldersUrl = `${PREFIX_DIRECTORY_SERVER_QUERIES}/v1/root-directories?${urlSearchParams}`;
+    return backendFetchJson(fetchRootFoldersUrl, {
+        method: 'get',
+        headers: { 'Content-Type': 'application/json' },
+    });
+}
+
+export function fetchDirectoryContent(directoryUuid: UUID, types?: string[]) {
```
Tristan-WorkGH, line 39:
```suggestion
export function fetchDirectoryContent(directoryUuid: UUID, types?: string[]): Promise<???> {
```
we don't want a `Promise<any>`.

src/services/directory.ts
```diff
@@ -0,0 +1,77 @@
...
+const PREFIX_DIRECTORY_SERVER_QUERIES =
+    import.meta.env.VITE_API_GATEWAY + '/directory';
+
+function getPathUrl(elementUuid: UUID) {
+    return `${PREFIX_DIRECTORY_SERVER_QUERIES}/v1/elements/${encodeURIComponent(
+        elementUuid
+    )}/path`;
+}
+
+export function fetchRootFolders(types: string[]) {
```
Tristan-WorkGH, line 24:
```suggestion
export function fetchRootFolders(types: string[]): Promise<???> {
```
we don't want a `Promise<any>`.

Tristan-WorkGH (changes requested):
part2

Have you look to a more update version of utilities function for rest base from gridadmin ?
  * [`parseError`](https://github.com/gridsuite/gridadmin-app/blob/664aee0be5529417fa7a1920fbcbeb750cd0fc5e/src/utils/api.ts#L40)
  * [`getErrorMessage`](https://github.com/gridsuite/gridadmin-app/blob/664aee0be5529417fa7a1920fbcbeb750cd0fc5e/src/utils/error.ts)
  * [api-rest.ts](https://github.com/gridsuite/gridadmin-app/blob/664aee0be5529417fa7a1920fbcbeb750cd0fc5e/src/utils/api-rest.ts)
  * [api-ws.ts](https://github.com/gridsuite/gridadmin-app/blob/664aee0be5529417fa7a1920fbcbeb750cd0fc5e/src/utils/api-ws.ts)

src/services/directory.ts
Tristan-WorkGH: yes, can be inline

src/components/DirectoryItemSelector/directory-item-selector.tsx
```diff
@@ -142,88 +133,70 @@ const DirectoryItemSelector: FunctionComponent<DirectoryItemSelectorProps> = ({
...
-                    });
+                            addToDirectory(nodeId, children);
+                        });
+                    } else {
+                        // update directory content
+                        addToDirectory(nodeId, childrenMatchedTypes);
+                    }
+                })
+                .catch((error) => {
+                    console.warn(
```
Tristan-WorkGH, line 194: only a console log?

src/index.ts
```diff
@@ -216,4 +216,5 @@ export {
 
 export { mergeSx } from './utils/styles';
 export { setCommonStore } from './redux/commonStore';
+export * from './services';
```
Tristan-WorkGH, line 219:
```suggestion
export * from './services';
export type * from './services';
```

src/services/metadata.ts
```diff
@@ -0,0 +1,39 @@
...
+    appColor: string;
+    hiddenInAppsMenu: boolean;
+    resources: unknown;
+}
+
+export interface StudyMetadata extends Metadata {
+    name: 'Study';
+    predefinedEquipmentProperties: {
+        [networkElementType: string]: PredefinedProperties;
+    };
```
Tristan-WorkGH, line 22: @TheMaskedTurtle a preference?
```suggestion
    predefinedEquipmentProperties: Record<string, PredefinedProperties>;
```

Tristan-WorkGH (changes requested):
part3

src/services/apps-metadata.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * Copyright (c) 2024, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+export type Url = string | URL;
+
+export function fetchEnv() {
```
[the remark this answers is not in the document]
Meklo: Issue is, it can't resolve `../../public/env.json` during compilation in commons-ui

src/services/utils.ts
Meklo: True, wasn't worth the pain to move in the end, I reverted it

src/services/utils.ts
```diff
@@ -80,3 +121,22 @@ const parseError = (text: string) => {
...
+    params: string[] = [],
+    paramName: string
+) => {
+    return new URLSearchParams(params.map((param) => [paramName, param]));
+};
+
+export const getWsBase = () =>
+    document.baseURI
+        .replace(/^http:\/\//, 'ws://')
+        .replace(/^https:\/\//, 'wss://');
```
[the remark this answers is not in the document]
Meklo: Done

src/services/utils.ts
```diff
@@ -12,11 +12,52 @@ export const backendFetch = (url: string, init: any, token?: string) => {
...
+export async function backendFetchFile(
+    url: string,
+    init?: any,
+    token?: string
+) {
+    return (await backendFetch(url, init, token)).blob();
+}
+
+export enum FileType {
+    ZIP,
```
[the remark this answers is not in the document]
Meklo: Done

src/services/metadata.ts
```diff
@@ -0,0 +1,39 @@
...
+    defaultCountry?: string;
+}
```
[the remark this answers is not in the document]
Meklo: Done

src/services/index.ts
```diff
@@ -0,0 +1,12 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+export * from './utils';
+export * from './apps-metadata';
+export * from './metadata';
+export * from './explore';
+export * from './directory';
+export * from './study';
```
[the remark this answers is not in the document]
Meklo: Done

src/services/utils.ts
Meklo: Done

src/services/metadata.ts
Meklo: Done

src/services/metadata.ts
Meklo: Done

src/index.ts
Meklo: Done

src/services/explore.ts
```diff
@@ -53,3 +57,40 @@ export function saveFilter(filter: any, name: string, token?: string) {
...
+    const elementTypesParams = getRequestParamFromList(
+        elementTypes,
+        'elementTypes'
+    );
+
+    const urlSearchParams = new URLSearchParams([
+        ...idsParams,
+        ...equipmentTypesParams,
+        ...elementTypesParams,
+    ]).toString();
```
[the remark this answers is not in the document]
Meklo: Based on https://stackoverflow.com/questions/45126076/how-can-i-combine-two-urlsearchparams
It seems to be possible to combine them this way :  `new URLSearchParams({
  ...Object.fromEntries(params1),
  ...Object.fromEntries(params2)
});`

src/components/DirectoryItemSelector/directory-item-selector.tsx
Meklo: How so ? Since it's in a catch bock I'd reckon a warn is warranted

src/services/directory.ts
Meklo: I've added `ElementAttributes` type which applies here

src/services/directory.ts
Meklo: I've added `ElementAttributes` type which applies here

src/services/directory.ts
Meklo: Done

src/services/directory.ts
Meklo: Done

src/services/utils.ts
TheMaskedTurtle: I like this solution https://stackoverflow.com/questions/41103360/how-to-use-fetch-in-typescript

src/services/utils.ts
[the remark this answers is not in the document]
Meklo: Had to rollback, it broke gridstudy

src/services/apps-metadata.ts
```diff
@@ -4,7 +4,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
+import { PredefinedProperties } from '../utils/types';
 
+// https://github.com/gridsuite/deployment/blob/main/docker-compose/docker-compose.base.yml
+// https://github.com/gridsuite/deployment/blob/main/k8s/resources/common/config/apps-metadata.json
+export type MetadataJson = MetadataCommon | MetadataStudy;
```
flomillot, line 11: Why `MetadataJson` and not just `Metadata` ? Every object in JS is a JSON anyway.

src/services/apps-metadata.ts
```diff
@@ -16,6 +20,7 @@ export type MetadataCommon = {
     url: Url;
     appColor: string;
     hiddenInAppsMenu: boolean;
+    defaultCountry?: string;
```
flomillot, line 23: It's only in Study metadata no ? Why in commons ?

src/services/directory.ts
```diff
@@ -36,24 +33,35 @@ export function fetchRootFolders(types: string[]) {
...
 
     // Add params to Url
     const urlSearchParams = getRequestParamFromList(
         types,
         'elementTypes'
     ).toString();
-    const fetchDirectoryContentUrl = `${PREFIX_DIRECTORY_SERVER_QUERIES}/v1/directories/${directoryUuid}/elements?${urlSearchParams}`;
+
+    let fetchDirectoryContentUrl = `${PREFIX_DIRECTORY_SERVER_QUERIES}/v1/directories/${directoryUuid}/elements`;
+    if (urlSearchParams.length > 0) {
```
flomillot, line 49:
```suggestion
    if (urlSearchParams) {
```
It's a string no ?

src/services/utils.ts
```diff
@@ -60,59 +67,44 @@ const downloadFile = (blob: Blob, filename: string, type?: FileType) => {
     document.body.removeChild(link);
 };
 
-const prepareRequest = (init: any, token?: string) => {
+function prepareRequest(init?: InitRequest, token?: Token): RequestInit {
```
flomillot, line 70: If we start to type the Token, it need to be changed everywhere

src/services/metadata.ts
flomillot: Is it exactly the same ?
If so, I prefer the first one, because the name can help to understand the object.

src/services/utils.ts
```diff
@@ -60,59 +67,44 @@ const downloadFile = (blob: Blob, filename: string, type?: FileType) => {
...
     if (!(typeof init === 'undefined' || typeof init === 'object')) {
         throw new TypeError(
-            'First argument of prepareRequest is not an object : ' + typeof init
+            `Argument 2 of backendFetch is not an object ${typeof init}`
```
flomillot, line 73: ?

src/services/utils.ts
```diff
@@ -60,59 +67,44 @@ const downloadFile = (blob: Blob, filename: string, type?: FileType) => {
...
-            customError = new Error(
-                errorName +
-                    errorJson.status +
-                    ' ' +
-                    errorJson.error +
-                    ', message : ' +
-                    errorJson.message
-            );
-            customError.status = errorJson.status;
+        if (errorJson?.status && errorJson?.error && errorJson?.message) {
```
flomillot, line 94: Actually the previous one is better because it only check errorJson once

src/services/utils.ts
```diff
@@ -60,59 +67,44 @@ const downloadFile = (blob: Blob, filename: string, type?: FileType) => {
...
+            ) as ErrorWithStatus;
+            error.status = errorJson.status;
         } else {
-            customError = new Error(
-                errorName +
-                    response.status +
-                    ' ' +
-                    response.statusText +
-                    ', message : ' +
-                    text
```
flomillot, line 68: So we don't want to display this anymore ?

flomillot (dismissed):

src/services/metadata.ts
TheMaskedTurtle: yes the record is better

src/services/directory.ts
Meklo: It's a string once toString has been called but has is it is an object like a Map, but I just tried out, if urlSearchParams is empty `if (urlSearchParams) {` returns true but `if (urlSearchParams.length > 0) {` doesn't work either, so I guess the correct condition would be `if (urlSearchParams.toString()) {`

src/services/apps-metadata.ts
Meklo: Fine by me

src/services/apps-metadata.ts
Meklo: Wasn't aware it's not a common one, I moved it to MetadataStudy

src/services/apps-metadata.ts
Tristan-WorkGH: because that is a file that we fetch, not an API of a server

src/services/utils.ts
Tristan-WorkGH: just a bad copy from another project I think

src/services/apps-metadata.ts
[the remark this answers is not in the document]
Tristan-WorkGH: Yes, that was to not have to manage manually the type definition manually.  
You just remove the link to the file in the definition and it's good.

src/services/utils.ts
Meklo: Restored it

src/services/utils.ts
flomillot: Nope it's the first argument, not the second

src/services/utils.ts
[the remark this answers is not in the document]
Tristan-WorkGH: ❓❕

src/services/apps-metadata.ts
flomillot: Not a good reason for me, we should not write "Json" at the end of a type because everything is a Json, whatever it come from, but maybe ask a TL to be sure

flomillot: If Metadata is not good, which I agree a bit, other possibility : 

- AppsMetadata
- CommonMetadata
- StudyMetadata
- etc...

For me it's even better.
@TheMaskedTurtle and the others, what do you think ?

src/services/metadata.ts
flomillot: I mean the other one x) don't you think networkElementType help to understand the object ?

src/services/directory.ts
flomillot: I don't get it, we need to call toString() two times ?

src/utils/ElementAttributes.ts
Tristan-WorkGH, line 1: If it's only to define a type, maybe not in `src/utils/` but in `src/types/` for example.

Also is it meant to be use multiple time or only once (actual case)? If the later I would move it where it's used.

src/services/utils.ts
```diff
@@ -7,71 +7,109 @@
...
+}
 
-const safeFetch = (url: string, initCopy: any) => {
+function safeFetch<T>(url: string, initCopy: any): Promise<T> {
     return fetch(url, initCopy).then((response) =>
         response.ok ? response : handleError(response)
-    );
-};
+    ) as Promise<T>;
+}
```
Tristan-WorkGH, line 87:
```suggestion
function safeFetch(url: string, initCopy: RequestInit): Promise<Response> {
    return fetch(url, initCopy).then((response) =>
        response.ok ? response : handleError(response)
    );
}
```
* `initCopy` is of type `RequestInit`
* `safeFetch` always return a `Response` or failed promise

src/services/utils.ts
Tristan-WorkGH: @Meklo
My remark wasn't for `backendFetch` but for `backendFetchText` + `backendFetchJson` + `backendFetchBlob` 😓

src/services/utils.ts
Tristan-WorkGH: > It's from the version of gridadmin, both are true in the end no ?

The message haven't been updated in gridadmin I remember.

src/services/utils.ts
Tristan-WorkGH: That was because we consider that as we only fetch from our servers and we know that they return errors either as details json or juste http status, we don't have case where we can have raw text.

On top of that, if we get for example a webpage (which isn't json), we don't want to display uncontrolled content.

src/services/utils.ts
Tristan-WorkGH: That's the idea

src/services/utils.ts
Meklo: Doesn't `backendFetch` also need to be parametrized tho ? Otherwise how can we cast `Promise<Response>` to the desired type, see functions in study service

src/services/utils.ts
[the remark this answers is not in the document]
Meklo: I must have badly replicated it, I retried and it's now fine

src/services/utils.ts
Meklo: Done

src/services/utils.ts
Meklo: Done

dbraquart (commented):
code: ok
tests: to be continued

Tristan-WorkGH (approved):
now assigned to me

PR #470, opened 2024-06-07
Tweak elements attributes and fix AG grid theming

Meklo:


src/utils/types.ts
```diff
@@ -40,6 +41,12 @@ export type ElementAttributes = {
     lastModifiedBy: string;
     children: any[];
     parentUuid: null | UUID;
+    specificMetadata: {
```
achour94, line 44: This is different for each element type

achour94: i suggest to put it as 'object' and create another PR to specify a type for each element like 'StudySpecificMetaData', 'FilterSpecificMetadata' ...

achour94 (approved):

klesaulnier (approved):

PR #473, opened 2024-06-13
Fix custom ag grid editing after sorting

achour94:


achour94:
another PR is created https://github.com/gridsuite/commons-ui/pull/476

PR #475, opened 2024-06-13
update hvdc contingency label

Mathieu-Deharbe:
partial revert from : 
https://github.com/gridsuite/commons-ui/pull/450


jamal-khey (approved):
Code OK, the label exist in grid-study

PR #476, opened 2024-06-14
Fix ag grid RHF update rows and disable move buttons when we have a sort

achour94:
**Issue:**
When a sort is applied in AG Grid, the values in React Hook Form (RHF) are not updated correctly. This causes editing to affect the wrong row, as the row index does not match the sorted order.

**Fix:**
In this PR, we fixed the issue by ensuring that the correct row index is retrieved based on the unique identifier 'AG_GRID_ROW_UUID'. This ensures that edits are applied to the correct row even after sorting.

Additionally, the row dragging feature in AG Grid is managed automatically (rowDragManaged is set to true), which disables dragging when sorting is applied ([AG Grid Row Dragging](https://www.ag-grid.com/react-data-grid/row-dragging/#managed-dragging)). To maintain consistent behavior with the buttons below the table, we have disabled the move buttons when sorting is applied. This ensures that row movement via buttons is also disabled when the grid is sorted, providing a consistent user experience.


src/components/inputs/react-hook-form/ag-grid-table/custom-ag-grid-table.tsx
```diff
@@ -227,6 +236,24 @@ export const CustomAgGridTable: FunctionComponent<CustomAgGridTableProps> = ({
...
+                return;
+            }
+            update(rowIndex, event.data);
+        },
+        [getIndex, update]
+    );
+
+    const onSortChanged = useCallback((event: SortChangedEvent) => {
+        const isAnycolumnhasSort = event?.api
+            ?.getColumnState()
```
Meklo, line 252: I think the optional chaining isn't necessary here since event can't be undefined

achour94: Done

Meklo (approved):

PR #479, opened 2024-06-18
Remove oauth implicit code and old authentication flag

Tristan-WorkGH:
Look with [hidden whitespace](./files?diff=split&w=1) to reduce indentations diff


src/utils/AuthService.ts
```diff
@@ -109,113 +112,50 @@ function initializeAuthenticationDev(
...
     dispatch: Dispatch<unknown>,
     isSilentRenew: boolean,
-    idpSettings: Promise<Response>,
+    idpSettingsFetcher: IdpSettingsGetter,
     validateUser: UserValidationFunc,
-    authorizationCodeFlowEnabled: boolean,
     isSigninCallback: boolean
 ) {
-    return idpSettings
-        .then((r) => r.json())
```
TheMaskedTurtle, line 126: Actually I'm not really sure this brings any value... What's better than before doing it in the apps and not here ?

TheMaskedTurtle (changes requested):

src/utils/AuthService.ts
Tristan-WorkGH: duplicated step in all fronts, and the same than why the api fetchers functions in apps do the `.json()` themselves

src/utils/AuthService.ts
```diff
@@ -77,24 +87,17 @@ function handleSigninSilent(
         if (user == null || getIdTokenExpiresIn(user) < 0) {
             return userManager.signinSilent().catch((error: Error) => {
                 dispatch(setShowAuthenticationRouterLogin(true));
-                const errorIssuerCodeFlow = isIssuerErrorForCodeFlow(error);
-                const errorIssuerImplicitFlow =
-                    error.message ===
-                    'authority mismatch on settings vs. signin state';
-                if (errorIssuerCodeFlow) {
+                if (isIssuerErrorForCodeFlow(error)) {
                     // Replacing authority for code flow only because it's done in the hash for implicit flow
```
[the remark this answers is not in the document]
Tristan-WorkGH: I had a little hesitation on this `if`, by the name did I need to remove all the block if it's only an error that can happen with code flow mode?

src/utils/AuthService.ts
Tristan-WorkGH: is it okay to do like that to keep compatibility for gridcapa?

src/utils/AuthService.ts
```diff
@@ -124,7 +123,12 @@ export async function initializeAuthenticationProd(
     validateUser: UserValidationFunc,
     isSigninCallback: boolean
 ) {
-    const idpSettings = await idpSettingsFetcher();
+    const idpSettings = await idpSettingsFetcher().then((r) =>
+        // backward compatibility with old callers
+        Object.prototype.toString.call(r) !== '[object Response]'
```
TheMaskedTurtle, line 128: Erf... I would rather avoid that kind of glue... Can we just go back to previous behavior ?

TheMaskedTurtle (changes requested):

src/utils/AuthService.ts
Tristan-WorkGH: okay

src/utils/AuthService.ts
```diff
@@ -383,9 +321,8 @@ function handleSigninCallback(
     userManagerInstance
         .signinRedirectCallback()
         .catch(function (e) {
-            if (isIssuerErrorForCodeFlow(e)) {
+            if (isIssuerError(e)) {
                 // Replacing authority for code flow only because it's done in the hash for implicit flow
```
TheMaskedTurtle, line 325: This can be removed too

src/utils/types.ts
```diff
@@ -24,6 +24,8 @@ import {
     VSC,
 } from './equipment-types';
 
+export type UrlString = `${string}://${string}` | `/${string}` | `./${string}`;
```
TheMaskedTurtle, line 27: I think this is not useful, it is not runtime validation and we don't use these defintion properties in the code (in parsing for example), so it is an overprecision that does not bring value.

src/utils/AuthService.ts
```diff
@@ -109,113 +111,49 @@ function initializeAuthenticationDev(
...
 }
 
 const accessTokenExpiringNotificationTime = 60; // seconds
 
-function initializeAuthenticationProd(
+export async function initializeAuthenticationProd(
     dispatch: Dispatch<unknown>,
     isSilentRenew: boolean,
-    idpSettings: Promise<Response>,
+    idpSettingsFetcher: IdpSettingsGetter,
```
TheMaskedTurtle, line 122: There is a naming mismatch that makes things confusing

TheMaskedTurtle (changes requested):

src/utils/AuthService.ts
Tristan-WorkGH: you mean between "fetcher" and "getter"? that synonyms for me 🤔

src/utils/types.ts
Tristan-WorkGH: It does have an effect with tests because they were failing after the first version, but yes it's limited in usage as IdP settings are runtime data.

But isn't better as documentation and validation, as it's extends by [`Url` type](https://github.com/gridsuite/commons-ui/blob/auth/remove_implicit_code/src/services/apps-metadata.ts#L11) after?

src/utils/AuthService.ts
TheMaskedTurtle: then choose one :grin:

src/utils/types.ts
TheMaskedTurtle: How could it make tests fail ? Which extension are you talking about ? I don't really see any benefit

TheMaskedTurtle (approved):

Tristan-WorkGH:
Waiting for gridcapa to migrate to code flow before merging.

PR #484, opened 2024-06-19
error message for invalid filter type

Mathieu-Deharbe:
Adds an error message when a user tries to open a filter with an invalid/obsolete equipment type.


jamal-khey (approved):
I noticed that we're handling a specific case here that seems to be very rare or might never happen. 

To simplify our codebase, have we considered removing all objects from the database that have this non-supported value instead? This approach could potentially eliminate the need for this check altogether, making our code cleaner and more efficient.
@etiennehomer  ?

Code OK
I tested by changing values in the db directly.

Mathieu-Deharbe:
Yes this error has very few chances to happen but the code may also be useful in order to check enum consistencies between the front and the back.

We could have removed the obsolete objects from the database of course, but this would mean destroying user data without any explication. Even worse if we take into account that those filters are directly used in some user config and computation parameters. Here they will get a nice error message and an opportunity to update their filters or remove them themselves.

PR #492, opened 2024-06-24
Change about dialog logo text

oualid-aloui-rte:


Tristan-WorkGH:
@oualid-aloui-rte did the solution #493 fulfill your need?

PR #505, opened 2024-07-01
Fix PR #493

Tristan-WorkGH:
Fix broken module name in 901ce6a67452abfd2fe239c3eb66359c45d211a5
* Have revert module name composition to before #493
* ~Simplify logo usage, so~
  - ~`<TopBar /*...*/ />` or `<TopBar logoAboutDialog={undefined} /*...*/ />`: give the "GridSuite" logo we have in GridSuite apps~
  - ~`<TopBar logoAboutDialog={null} /*...*/ />`: give the same logo than in top-bar (reuse the same instance internally)~
  - ~`<TopBar logoAboutDialog={anything} /*...*/ />`: will use the _anything_ ReactNode in place of the logo~

### DIFF
--- a/src/components/TopBar/TopBar.tsx
+++ b/src/components/TopBar/TopBar.tsx
@@ -663,11 +663,7 @@ const TopBar = ({
                     appLicense={appLicense}
                     globalVersionPromise={globalVersionPromise}
                     additionalModulesPromise={additionalModulesPromise}
-                    logo={
-                        logoAboutDialog === null
-                            ? logo_clickable
-                            : logoAboutDialog
-                    }
+                    logo={logoAboutDialog}
                 />
             </Toolbar>
         </AppBar>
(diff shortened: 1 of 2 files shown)


src/components/TopBar/TopBar.tsx
```diff
@@ -663,7 +663,11 @@ const TopBar = ({
                     appLicense={appLicense}
                     globalVersionPromise={globalVersionPromise}
                     additionalModulesPromise={additionalModulesPromise}
-                    logo={logoAboutDialog}
+                    logo={
```
antoinebhs, line 666: Keep the behavior that you implemented. Fix only the app name in the details section.

antoinebhs (changes requested):

antoinebhs (approved):
Code OK
Test OK

PR #518, opened 2024-07-11
chore: migrate from jest to vitest

TheMaskedTurtle:


vitest.config.mts
```diff
@@ -0,0 +1,22 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/// <reference types="vitest" />
+import { defineConfig } from 'vite';
+import react from '@vitejs/plugin-react';
+
+export default defineConfig({
+    plugins: [react()],
```
jonenst, line 13: do we really need to redefine a vitestconfig file ?
"f you are using Vite and have a vite.config file, Vitest will read it to match with the plugins and setup as your Vite app. If you want to have a different configuration for testing [...] you could [...] create vitest.config.ts"

vitest.setup.mts
```diff
@@ -5,7 +5,10 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { TextEncoder, TextDecoder } from 'util';
+import { afterEach } from 'vitest';
+import { cleanup } from '@testing-library/react';
```
jonenst, line 9: there was already a PR to add @testing-library/react https://github.com/gridsuite/commons-ui/pull/216/files?w=1 it simplifies the code we should use it if we add testing-library

vitest.config.mts
capyq: on other project, we set this directly on vite.config file with those parameters 
`test: {
    globals: true,
    environment: 'jsdom',
    setupFiles: './src/testSetup.ts',
  },`
and the setup file, define a default config, like after `afterEach(() => { cleanup() });`

PR #524, opened 2024-07-17
add category hvdc for expert filters

EtienneLt:
linked to https://github.com/gridsuite/gridexplore-app/pull/473 and https://github.com/gridsuite/filter/pull/41


src/components/filter/expert/expert-filter-constants.ts
```diff
@@ -860,4 +899,27 @@ export const fields: Record<string, Field[]> = {
         FIELDS_OPTIONS.SUBSTATION_PROPERTY,
         FIELDS_OPTIONS.VOLTAGE_LEVEL_PROPERTY,
     ],
+    HVDC_LINE: [
```
thangqp, line 902: I don't see declaration for NOMINAL_VOLTAGE in the fields
see https://github.com/gridsuite/filter/pull/41/files#diff-a6fa1d8bbf07e8485a7dc4932ca9a10bb0ef0d7af1e0afcbef77f88b542c4fb6R72

But i don't think we can reuse NOMINAL_VOLTAGE because this is voltage of the line so it should be another label

src/components/filter/expert/expert-filter-constants.ts
```diff
@@ -674,6 +695,42 @@ export const FIELDS_OPTIONS = {
...
+    },
+    CONVERTER_STATION_ID_2: {
+        name: FieldType.CONVERTER_STATION_ID_2,
+        label: 'converterStationId2',
+        dataType: DataType.STRING,
+    },
+    CONVERTERS_MODE: {
+        name: FieldType.CONVERTERS_MODE,
+        label: 'convertersMode',
+        dataType: DataType.ENUM,
```
thangqp, line 711: Even this field is declared as enum but translation is not done

Capture d’écran du 2024-07-19 15-13-06

I have a suggestion in in PR: https://github.com/gridsuite/commons-ui/pull/530

thangqp (approved):

PR #542, opened 2024-08-01
rework : split AuthenticationRouter into atomic component

capyq:
Extract subcomponents into their own files.


src/components/AuthenticationRouter/AuthenticationRouter.tsx
```diff
@@ -77,10 +45,10 @@ function AuthenticationRouter({
             alignItems="center"
             direction="column"
         >
-            {userManager.error !== null && (
+            {userManager.error && (
```
Tristan-WorkGH, line 48: We can have an empty string if we haven't retrieve the error message.

src/components/AuthenticationRouter/authenticationType.ts
```diff
@@ -0,0 +1,33 @@
...
+    logoutError?: { error: Error };
+    unauthorizedUserInfo?: string;
+};
+
+export type UserManagerState = {
+    instance: UserManager | null;
+    error: string | null;
+};
+
+export interface AuthenticationRouterProps {
```
Tristan-WorkGH, line 25: Don't we keep the props near the component?

Tristan-WorkGH (commented):
We don't touch `src/utils/AuthService`?

capyq:
> We don't touch `src/utils/AuthService`?

It's a minimal rework of the component; the service will be reworked later.

capyq:
@TheMaskedTurtle, rebase done.

TheMaskedTurtle (approved):

PR #546, opened 2024-08-06
fix translation for hvdc converter

EtienneLt:
get the same name for hvdc converter mode from network modifications


EtienneLt:
I'm not sure but I think these labels should be removed from gridstudy-app and gridexplore-app

PR #549, opened 2024-08-08
WIP: Fix null exception in directory item selector

jamal-khey:
This pull request fixes an issue where the directory item selector's `updatedTree` function could throw a null exception. 

this issu happen in modification by formula , when selecting a filter and then moving it to another directory. when the user re-click on the chip of the filter , the derictory can not be selected


AAJELLAL (commented):
As we have seen together, there are cases where behavior is strange. It's best to ask the PO about the reproduction scenario and the various test cases.

src/components/DirectoryItemSelector/directory-item-selector.tsx
```diff
@@ -77,7 +77,7 @@ function updatedTree(prevRoots: any[], prevMap: any, nodeId: UUID | null, childr
                 return pn;
             }
             if (pn.parentUuid !== nodeId) {
-                console.warn(`reparent ${pn.parentUuid} -> ${nodeId}`);
+                console.warn(`reparent ${pn?.parentUuid} -> ${nodeId}`);
```
capyq, line 80: I don't understant how it fix null exception. 

if you optional chainning is the fix, that mean that "pn" is undefined.
and line 79 if "pn" is undefined you will get the same null exception.

line 68 if "pn" is undefined we return a value, so line 80 pn couldn't be undefined.

capyq (changes requested):

src/components/DirectoryItemSelector/directory-item-selector.tsx
jamal-khey: I apologize for the confusion. The title of the pull request was a mistake, and the bug is not resolved yet. I will investigate further to address the null exception issue properly.

PR #553, opened 2024-08-19
Add missing ElementType and update filterCreation props interface

ayolab:


achour94 (approved):

src/components/filter/filter-creation-dialog.tsx
```diff
@@ -54,7 +54,7 @@ const formSchema = yup
 
 export interface FilterCreationDialogProps {
     open: boolean;
-    onClose: () => void;
+    onClose: (e?: any, nextSelectedDirectoryId?: string | null) => void;
```
capyq, line 57: Never use 'any' prefer 'unknown' instead

achour94 (approved):

PR #562, opened 2024-08-23
Draft:Rework/architecture barrel

capyq:
PR to clean index file, fixing duplication. 

draft => "Equipment" is exported twice with differents definition
 => We need to search where there used to pick one


src/utils/types/types.ts
```diff
@@ -47,26 +30,23 @@ export type ElementAttributes = {
...
+//     | typeof SVC
+//     | typeof DanglingLine
+//     | typeof LCC
+//     | typeof VSC
+//     | typeof Hvdc
+//     | typeof BusBar
+//     | typeof TwoWindingTransfo
+//     | typeof ThreeWindingTransfo
+//     | typeof ShuntCompensator
+//     | typeof VoltageLevel;
```
TheMaskedTurtle, line 49: I checked about it, it's unused, I will do another PR to clean that :)

TheMaskedTurtle (approved):

PR #568, opened 2024-09-04
Fix: add sx props to list and container in checkbox list

YenguiSeddik:


src/components/CheckBoxList/check-box-list-type.ts
```diff
@@ -15,6 +15,8 @@ export interface CheckBoxListItemSxProps {
     checkboxList?: SxProps;
     checkboxButton?: SxProps;
     checkbox?: SxProps;
+    dragAndDropContainer?: SxProps;
+    list?: SxProps;
```
sBouzols, line 19: Rename it checkboxList and rename existing `checkboxList` to `checkboxListItem` which is used by items `CheckboxListItem` or `DraggableCheckBoxListItem`

YenguiSeddik: done

sBouzols (approved):

PR #573, opened 2024-09-11
improve filter ihm

Mathieu-Deharbe:
The point is fore the filter ceration/update ihm to have : 
- a fixed height (almost full height screen (95% actually))
- an unscrollable header composed of multiple component from various files (but with a scrollable lower part)

### DIFF
--- a/src/components/filter/explicitNaming/ExplicitNamingFilterForm.tsx
+++ b/src/components/filter/explicitNaming/ExplicitNamingFilterForm.tsx
@@ -239,32 +239,30 @@ export function ExplicitNamingFilterForm({ sourceFilterForExplicitNamingConversi
                 )}
             </Box>
             {watchEquipmentType && (
-                <Box sx={unscrollableDialogStyles.scrollableContent}>
-                    <CustomAgGridTable
-                        name={FILTER_EQUIPMENTS_ATTRIBUTES}
-                        columnDefs={columnDefs}
-                        defaultColDef={defaultColDef}
-                        makeDefaultRowData={makeDefaultRowData}
-                        pagination
-                        paginationPageSize={100}
-                        suppressRowClickSelection
-                        alwaysShowVerticalScroll
-                        stopEditingWhenCellsLoseFocus
-                        csvProps={{
-                            fileName: intl.formatMessage({
-                                id: 'filterCsvFileName',
-                            }),
-                            fileHeaders: csvFileHeaders,
-                            getDataFromCsv: getDataFromCsvFile,
-                        }}
-                        cssProps={{
-                            '& .ag-root-wrapper-body': {
-                                maxHeight: '380px',
-                                height: '380px',
-                            },
-                        }}
-                    />
-                </Box>
+                <CustomAgGridTable
+                    name={FILTER_EQUIPMENTS_ATTRIBUTES}
+                    columnDefs={columnDefs}
+                    defaultColDef={defaultColDef}
+                    makeDefaultRowData={makeDefaultRowData}
+                    pagination
+                    paginationPageSize={100}
+                    suppressRowClickSelection
+                    alwaysShowVerticalScroll
+                    stopEditingWhenCellsLoseFocus
+                    csvProps={{
+                        fileName: intl.formatMessage({
+                            id: 'filterCsvFileName',
+                        }),
+                        fileHeaders: csvFileHeaders,
+                        getDataFromCsv: getDataFromCsvFile,
+                    }}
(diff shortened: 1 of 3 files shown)


src/components/filter/criteriaBased/CriteriaBasedForm.tsx
```diff
@@ -58,19 +79,26 @@ function CriteriaBasedForm({ equipments, defaultValues }: CriteriaBasedFormProps
...
-                    );
-                })}
-        </Grid>
+            </Box>
+            <Box sx={styles.ScrollableContainer}>
+                <Box sx={styles.ScrollableContent}>
+                    <Grid container spacing={2}>
+                        {watchEquipmentType &&
+                            equipments[watchEquipmentType] &&
+                            equipments[watchEquipmentType].fields.map((equipment: any, index: number) => {
```
ghazwarhili, line 88: any is not accepted

src/components/dialogs/customMuiDialog/CustomMuiDialog.tsx
```diff
@@ -40,6 +41,18 @@ const styles = {
             margin: 'auto',
         },
     },
+    dialogPaperFullHeight: {
+        '.MuiDialog-paper': {
+            width: 'auto',
+            minWidth: '1100px',
+            margin: 'auto',
+            height: '95vh',
```
basseche, line 49: - Height depends on navigator window height (i thought in specs that it should be fix).
- when i reduce navigator window height i can't access content any more there should be a scrollbar no ?
Capture d’écran de 2024-10-09 11-29-42

- when i reduce navigator window width, i can't see anything
Capture d’écran de 2024-10-09 11-33-07

thangqp: To be careful, this component is used in different creation dialogues, e.g create study, etc.

Mathieu-Deharbe: **To Thang** : don't worry I added a param for my own version of this dialog box : other uses of this dialog box won't be affected.

**To Bassel** : I spoke with Pauline and Stéphane : The viewport of the user should be at least 1024x880 : we don't care about smaller resolution. minWidth 1100 px is a bit too big though. I will reduce it a little bit (only for my own version).

And what specs are you talking about ? Can you share ? I don't know any reason not to use window height percentage. This is used elsewhere (for the transformers modification I think).

Mathieu-Deharbe: minWidth changed.

src/components/filter/criteriaBased/CriteriaBasedForm.tsx
thangqp: we can profit this PR to declare equipment here as FormField

Mathieu-Deharbe: Ok, typed s a FormField0

src/components/filter/explicitNaming/ExplicitNamingFilterForm.tsx
```diff
@@ -32,6 +32,24 @@ import { EquipmentType } from '../../../utils/types/equipmentType';
 
 export const FILTER_EQUIPMENTS_ATTRIBUTES = 'filterEquipmentsAttributes';
 
+const styles = {
+    ScrollableContainer: {
```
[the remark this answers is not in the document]
Mathieu-Deharbe: Changed (even if most of those are probably going to change or disappear).

src/components/filter/explicitNaming/ExplicitNamingFilterForm.tsx
```diff
@@ -236,34 +254,39 @@ export function ExplicitNamingFilterForm({ sourceFilterForExplicitNamingConversi
...
+                        />
+                    </Box>
+                </Box>
             )}
-        </Grid>
+        </>
     );
 }
+
+export default ExplicitNamingFilterForm;
```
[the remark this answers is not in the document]
Mathieu-Deharbe: Removed.

src/components/dialogs/customMuiDialog/CustomMuiDialog.tsx
Mathieu-Deharbe: Ok I did.
But not sure that this is the right move : it may be better to have small margins on both sides at all times.

basseche (approved):

PR #584, opened 2024-09-20
tests: add card error boundary tests => After tests/commonUtils

capyq:


package.json
```diff
@@ -30,6 +30,7 @@
     "dependencies": {
         "@react-querybuilder/dnd": "^7.2.0",
         "@react-querybuilder/material": "^7.2.0",
+        "@testing-library/user-event": "^14.5.2",
```
jonenst, line 33: should be a devdependency ?

jonenst (approved):

PR #589, opened 2024-09-25
Reduce list height and add autocomplete to field select component

basseche:


src/components/inputs/reactQueryBuilder/FieldSelector.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * Copyright (c) 2023, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { FullOption, toFlatOptionArray, ValueSelectorProps } from 'react-querybuilder';
+import { Autocomplete, TextField } from '@mui/material';
+
+function FieldSelector(props: ValueSelectorProps) {
```
Mathieu-Deharbe, line 11:
```suggestion
function FieldSelector(props: ReadOnly<ValueSelectorProps>) {
```
There is new control about this in sonar now. But it looks a good idea anyway.

basseche: yes i saw that too. Thanks i will modify this.

src/components/inputs/reactQueryBuilder/FieldSelector.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { FullOption, toFlatOptionArray, ValueSelectorProps } from 'react-querybuilder';
+import { Autocomplete, TextField } from '@mui/material';
+
+function FieldSelector(props: Readonly<ValueSelectorProps>) {
+    const { options, className, value, disabled, handleOnChange } = props;
```
Mathieu-Deharbe, line 12:
```suggestion
function FieldSelector({ options, className, value, disabled, handleOnChange }: Readonly<ValueSelectorProps>) {
```
New cooler way to do the same thing. I think @TheMaskedTurtle prefer like this but not sure at all. So do as you wish.

Mathieu-Deharbe (changes requested):
Not a big deal for me but you don't follow the exact specifications : 

"_La valeur retenue doit nécessairement être dans la liste : en validant sa saisie avec la touche Entrée la première valeur de la liste est automatiquement sélectionnée. Et à défaut le champ est laissé vide, qui est un état invalide ne permettant pas de valider la création/édition du filtre."_

The "enter" key doesn't work like this in any of those 2 cases.

I also think you should use the same interface for the right side value selector (even if for now they don't hold much values, they might, especially properties). See 'Hydro" here : 

image

src/components/inputs/reactQueryBuilder/FieldSelector.tsx
thangqp: Agreed with Mathieu, now destructuring in the signature is in convention except some particular case.
In addition, I dont see the befifice of using Readonly<T> for props if we destruct in the signature

thangqp:
> Not a big deal for me but you don't follow the exact specifications :
> 
> "_La valeur retenue doit nécessairement être dans la liste : en validant sa saisie avec la touche Entrée la première valeur de la liste est automatiquement sélectionnée. Et à défaut le champ est laissé vide, qui est un état invalide ne permettant pas de valider la création/édition du filtre."_
> 
> The "enter" key doesn't work like this in any of those 2 cases.
> 
> I also think you should use the same interface for the right side value selector (even if for now they don't hold much values, they might later, especially _properties_). See 'Hydro" here :
> 
> image

The first issue is related to : https://github.com/gridsuite/commons-ui/blob/main/src/components/filter/expert/ExpertFilterForm.tsx#L79-L81 and each time add a new rule may be related to https://react-querybuilder.js.org/docs/components/querybuilder#autoselectfield, but it will add a "-----" option that I don't think it fit to the demand of PO.

The second issue is just add autoHighlight to Autocomplete

thangqp (approved):

Mathieu-Deharbe (approved):

PR #598, opened 2024-10-02
split in several requests getMetadata when there are too many ids

EtienneLt:
For large number of ids, the request does not pass because the headers are to large.
The request will be cut in several smaller


capyq (approved):

src/services/explore.ts
```diff
@@ -49,33 +49,35 @@ export function saveFilter(filter: any, name: string, token?: string) {
...
-
-    const url = `${PREFIX_EXPLORE_SERVER_QUERIES}/v1/explore/elements/metadata?${urlSearchParams}`;
-    console.debug(url);
-    return backendFetchJson(url, {
-        method: 'get',
-        headers: { 'Content-Type': 'application/json' },
-    });
+        const url = `${PREFIX_EXPLORE_SERVER_QUERIES}/v1/explore/elements/metadata?${urlSearchParams}`;
+        // eslint-disable-next-line no-await-in-loop
+        const result = await backendFetchJson(url, {
```
sBouzols, line 76: Do we need to wait for one request for one chunk before sending the next one ?

src/services/explore.ts
```diff
@@ -49,33 +49,35 @@ export function saveFilter(filter: any, name: string, token?: string) {
     );
 }
 
-export function fetchElementsInfos(
+export async function fetchElementsInfos(
```
sBouzols, line 52: Why using the async / await syntax here ? Every functions un services uses Promise format.
I'm sure it's not necessary here. A Promise.all should do the trick if it's acceptable to send multiple paralell requests and that the order of ElementsInfos doesn't matter.

Mathieu-Deharbe (commented):
Tested, OK.

Waiting for a confirmation on this : https://github.com/gridsuite/gridexplore-app/pull/534#discussion_r1793465303

sBouzols (approved):

PR #602, opened 2024-10-08
add isc filter criterium for voltage levels

EtienneLt:
Add new label for expert filter 
it allows to select max and min isc criterium on voltage levels


basseche (approved):

basseche (changes requested):
When i try to save an expert filter it doesn't work and i have the following error :

Capture d’écran de 2024-10-10 14-34-48

Mathieu-Deharbe (commented):
Personally I can save the filters using ICC min or ICC max. Maybe Bassel misused the lib ?

I can't test much though, because there are no ICC mi or ICC max in my test cases.

EtienneLt:
> Personally I can save the filters using ICC min or ICC max. Maybe Bassel misused the lib ?
> 
> I can't test much though, because there are no ICC mi or ICC max in my test cases.

I can send you a case

basseche:
Ok no problem if it works for @Mathieu-Deharbe  maybe i missed the lib.

src/components/filter/expert/expertFilterConstants.ts
```diff
@@ -1041,6 +1041,16 @@ export const FIELDS_OPTIONS = {
...
+    SHORT_CIRCUIT_CURRENT_LOW_LIMIT: {
+        name: FieldType.SHORT_CIRCUIT_CURRENT_LOW_LIMIT,
+        label: 'shortCircuitLowCurrentLimit',
+        dataType: DataType.NUMBER,
+    },
+    SHORT_CIRCUIT_CURRENT_HIGH_LIMIT: {
+        name: FieldType.SHORT_CIRCUIT_CURRENT_HIGH_LIMIT,
+        label: 'shortCircuitHighCurrentLimit',
+        dataType: DataType.NUMBER,
+    },
```
Mathieu-Deharbe, line 1053: In the previous enum value using `DataType.NUMBER`, like `DC_NOMINAL_VOLTAGE`, there was `inputType: 'number',` as well. I don't know the reason but maybe it is important...?

src/translations/en/filterExpertEn.ts
```diff
@@ -188,4 +188,6 @@ export const filterExpertEn = {
     YupNotTypeDefault: 'Field value format is incorrect',
     changeOperatorMessage:
         'The operator will be changed and will be applied to all the rules already created in the group.',
+    shortCircuitLowCurrentLimit: 'Low short-circuit current limit',
+    shortCircuitHighCurrentLimit: 'High short-circuit current limit',
```
Mathieu-Deharbe, line 192: You should add the units, just like for `dcNominalVoltage`.

Mathieu-Deharbe, line 192: Not important but why do you reorganize the order of the words for the key ? (and for the enums everywhere)

Mathieu-Deharbe (changes requested):
Only the unit part is important. Do as you wish with the rest.

Mathieu-Deharbe (approved):

basseche (approved):

EtienneLt:
filter library must be released first before merging

PR #616, opened 2024-10-16
disable cgm-with-subnetworks option

basseche:
We only added parameters in FlatParameters dialog. The change is made directly in gridExplore. see https://github.com/gridsuite/gridexplore-app/pull/544


etiennehomer (changes requested):
FlatParameters is a generic component. You cannot put some code that is so specific to import parameters. 
 You can make the API evolute with a list of params to disable. And in gridexplore, you give CGM_WITH_SUBNETWORKS_OP to this list

basseche:
> FlatParameters is a generic component. You cannot put some code that is so specific to import parameters. You can make the API evolute with a list of params to disable. And in gridexplore, you give CGM_WITH_SUBNETWORKS_OP to this list

Ok Changed

src/components/flatParameters/FlatParameters.tsx
```diff
@@ -264,7 +267,13 @@ export function FlatParameters({
         const fieldValue = mixInitAndDefault(param);
         switch (param.type) {
             case 'BOOLEAN':
-                return <Switch checked={!!fieldValue} onChange={(e) => onFieldChange(e.target.checked, param)} />;
+                return (
+                    <Switch
+                        checked={!!fieldValue}
+                        onChange={(e) => onFieldChange(e.target.checked, param)}
+                        disabled={disabledParams !== undefined && disabledParams.indexOf(param.name) >= 0}
```
thangqp, line 274:
```suggestion
                        disabled={disabledParams?.indexOf(param.name) >= 0}
```
It is better to use JS syntax..

src/components/flatParameters/FlatParameters.tsx
```diff
@@ -424,7 +433,11 @@ export function FlatParameters({
         <List sx={styles.paramList}>
             {paramsAsArray.map((param, index) => (
                 <Fragment key={param.name}>
-                    <ListItem sx={styles.paramListItem}>
+                    <ListItemButton
+                        sx={styles.paramListItem}
+                        disabled={disabledParams !== undefined && disabledParams.indexOf(param.name) >= 0}
```
thangqp, line 438:
```suggestion
                        disabled={disabledParams?.indexOf(param.name) >= 0}
```
Use JS syntax

thangqp: I don't understand why we have double checks, one as this level, and another in renderField?

src/components/flatParameters/FlatParameters.tsx
```diff
@@ -12,7 +12,7 @@ import {
     Divider,
     IconButton,
     List,
-    ListItem,
+    ListItemButton,
```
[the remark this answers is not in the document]
thangqp: Same question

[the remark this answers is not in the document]
basseche: Because disabled is deprecated in ListItemBaseProps 
Capture d’écran du 2024-10-21 11-30-09

src/components/flatParameters/FlatParameters.tsx
etiennehomer: Yes the one in renderField() is to removed

src/components/flatParameters/FlatParameters.tsx
basseche: it doesn't work !

src/components/flatParameters/FlatParameters.tsx
etiennehomer: The switch is in the ListItemButton. It's enough to disable the ListItemButton

basseche: ok solved

PR #630, opened 2024-11-05
Fix item selection when clicking on chips

YenguiSeddik:


src/components/directoryItemSelector/DirectoryItemSelector.tsx
```diff
@@ -273,19 +273,26 @@ export function DirectoryItemSelector({
     );
 
     useEffect(() => {
-        if (open) {
-            updateRootDirectories();
+        if (open && expanded) {
             if (expanded) {
```
souissimai, line 277: this should be always true

YenguiSeddik: done

jamal-khey (approved):
looks good to me

souissimai (approved):

PR #634, opened 2024-11-07
Refactor SnackbarProvider: clear/close button component

sBouzols:
Use a IconButton component for the clear button instead of the raw ✖ button text

This fix icon centering and size


YenguiSeddik (approved):

jamal-khey (approved):
code and test Ok.

PR #637, opened 2024-11-14
Fix TextField adornments warnings

Meklo:
Prevents the following warnings :

    Warning: React does not recognize the `adornmentText` prop on a DOM element.
    Warning: React does not recognize the `adornmentPosition` prop on a DOM element.

By not passing those props only when `Field` is a `TextField` instead of a `TextFieldWithAdornment`


dbraquart (approved):
ok "adornment" warnings are gone

PR #639, opened 2024-11-18
two windings and three windings transformers have only one site in ex…

basseche:
P.S : we don't migrate old filters with mistakes on sites


src/components/filter/expert/expertFilterConstants.ts
```diff
@@ -851,13 +851,6 @@ export const FIELDS_OPTIONS = {
         valueEditorType: 'select',
         defaultValue: '',
     },
-    SUBSTATION_PROPERTY_3: {
-        name: FieldType.SUBSTATION_PROPERTY_3,
-        label: 'substationProperty3Twt',
```
Mathieu-Deharbe, line 856: I am quite sure that you can remove the associated loc : 
* substationProperty3Twt: 'Substation property 3',
* substationProperty3Twt: 'Propriété site 3',

Mathieu-Deharbe (changes requested):
I tried to apply a filter using SUBSTATION_PROPERTIES and I got this error : 

image

I am afraid that SUBSTATION_PROPERTIES and SUBSTATION_PROPERTIES_1 are different in the back and that you might have to modify it there.

You'll probably also have to "migrate" the old SUBSTATION_PROPERTIES_1 into SUBSTATION_PROPERTIES.

src/components/filter/expert/expertFilterConstants.ts
EtienneLt: in both translations

basseche: done

Mathieu-Deharbe (approved):

EtienneLt (approved):

PR #669, opened 2025-01-16
change rule label

EtienneLt:


Mathieu-Deharbe (changes requested):
image

You should change the error message containing "rule" as well : betweenRule, emptyGroup, changeOperatorMessage

src/translations/en/filterEn.ts
```diff
@@ -23,8 +23,8 @@ export const filterEn = {
     emptyRule: 'Filter contains an empty field',
     incorrectRule: 'Filter contains an incorrect field',
     obsoleteFilter: 'This filter is no longer supported. Please remove it or change its equipment type.',
-    betweenRule: "Left value of 'between' rule have to be lower than the right value",
-    emptyGroup: 'Filter contains an empty group. Consider removing it or adding rules to this group',
+    betweenRule: "Left value of 'between' criterion have to be lower than the right value",
```
Mathieu-Deharbe, line 26: Not your fault but I can't unsee it ^^
```suggestion
    betweenRule: "Left value of 'between' criterion has to be lower than the right value",
```

Mathieu-Deharbe (approved):

PR #674, opened 2025-01-22
Improve gui criteria filter

basseche:


Mathieu-Deharbe (commented):
Not important at all but when you click on "is / est" (which cannot be modified) it remains selected  in a brightest color. Just like if it was editable.

image

src/hooks/index.ts
```diff
@@ -15,3 +15,4 @@ export * from './usePredefinedProperties';
 export * from './usePrevious';
 export * from './useSnackMessage';
 export * from './useFormatLabelWithUnit';
+export * from './useSelectAppearance';
```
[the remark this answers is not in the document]
basseche: we never know !

Mathieu-Deharbe (approved):

src/components/inputs/reactQueryBuilder/PropertyValueEditor.tsx
```diff
@@ -103,19 +127,26 @@ export function PropertyValueEditor(props: ExpertFilterPropertyProps) {
...
+                            placeholder={propertyValues?.length > 0 ? '' : intl.formatMessage({ id: 'valuesList' })}
+                        />
+                    )}
                     freeSolo
                     autoSelect
                     onChange={(event, value) => {
                         onChange(FieldConstants.PROPERTY_VALUES, value);
                     }}
                     size="small"
+                    filterOptions={(options, params) => filterList(options, params)}
```
thangqp, line 149: filterList is already a memorized function.. it can be used directly.. 
we can rename filterList to filterOptions, same as props name

src/hooks/useSelectAppearance.ts
```diff
@@ -0,0 +1,28 @@
...
+                boxShadow: 'none',
+                '.MuiOutlinedInput-notchedOutline': { border: 'none' },
+                pointerEvents: 'none',
+                border: 'none',
+            },
+            readOnly: true,
+            disableUnderline: true,
+        };
+    }
+    return {};
```
thangqp, line 27: a custom hook usually is used with a react hook for memorized, that why we use 'useXXX'.. we can use 'useMemo' to return a memorized value.
In addition we can add a comment to explain the whole props constant, for example // disable edition, outline of a select

src/components/inputs/reactQueryBuilder/PropertyValueEditor.tsx
```diff
@@ -70,8 +71,29 @@ export function PropertyValueEditor(props: ExpertFilterPropertyProps) {
         [valueEditorProps]
     );
 
+    const filter = useMemo(() => {
+        return createFilterOptions<string>();
+    }, []);
```
thangqp, line 76: can call createFilterOptions<string>() directly in the below useCallback to avoid unnecessary chaining dependencies

src/components/inputs/reactQueryBuilder/PropertyValueEditor.tsx
```diff
@@ -70,8 +71,29 @@ export function PropertyValueEditor(props: ExpertFilterPropertyProps) {
...
+
+    const filterList = useCallback(
+        (options: string[], params: FilterOptionsState<string>) => {
+            let filtered = filter(options, params);
```
thangqp, line 80:
```suggestion
            const fillterOptions = createFilterOptions<string>();
            let filteredOptions = fillterOptions(options, params);
```

thangqp, line 80:
```suggestion
            const filtered = filter(options, params);
```

src/components/inputs/reactQueryBuilder/PropertyValueEditor.tsx
```diff
@@ -70,8 +71,29 @@ export function PropertyValueEditor(props: ExpertFilterPropertyProps) {
...
+            const { inputValue } = params;
+
+            const isExisting = options.some((option) => inputValue === option);
+            if (isExisting && options.length === 1 && options[0] === inputValue) {
+                // exact match : nothing to show
+                filtered = [];
```
thangqp, line 86:
```suggestion
                return [];
```

src/components/inputs/reactQueryBuilder/PropertyValueEditor.tsx
```diff
@@ -70,8 +71,29 @@ export function PropertyValueEditor(props: ExpertFilterPropertyProps) {
...
+            } else if (inputValue !== '' && !isExisting) {
```
thangqp, line 87:
```suggestion
            } 
            
            if (inputValue !== '' && !isExisting) {
```

PR #684, opened 2025-02-05
seperate the two types of HVDC station

ghazwarhili:


src/utils/types/extendedEquipmentType.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { EQUIPMENT_TYPE, EquipmentType } from './equipmentType';
+
+export const ExtendedEquipmentType = {
+    ...EquipmentType,
+    HVDC_LINE_LCC: 'HVDC_LINE_LCC',
+    HVDC_LINE_VSC: 'HVDC_LINE_VSC',
+} as const;
+
+export const EQUIPMENT_INDEXED_TYPE: Record<
```
thangqp, line 16: I don't see the usage of this object, it should replace the usage of EQUIPMENT_TYPE some where in commons-ui?

thangqp: may be here: https://github.com/gridsuite/commons-ui/blob/main/src/components/elementSearch/elementItem/EquipmentItem.tsx#L60

thangqp: The second usage is at : https://github.com/gridsuite/commons-ui/blob/main/src/hooks/useModificationLabelComputer.tsx#L73
But I think there a mistake, should be EquipmentType.HVDC_LINE instead of EQUIPMENT_TYPE.HVDC_LINE which is an object (never compare two objects in JS)

PR #688, opened 2025-02-10
Naming filter reduce selection col width

basseche:


Mathieu-Deharbe (changes requested):
Your change applies also to the contingency list.
Before : 

image

Now : 

image

I think you should update the contingency list css  in order to avoid those very small column : they probably need a 'flex: 1' somewhere in order to take the whole width.

basseche:
> Your change applies also to the contingency list. Before :
> 
> image
> 
> Now :
> 
> image
> 
> I think you should update the contingency list css in order to avoid those very small column : they probably need a kind of 'flex: 1' or whatever somewhere in order to take the whole width.

Solved in PR : 

https://github.com/gridsuite/gridexplore-app/pull/611

since it is inside GridExplore

Mathieu-Deharbe (approved):

PR #694, opened 2025-02-17
filter ihm improvements

Mathieu-Deharbe:


basseche:
The size of field name is changed everywhere in a strange way : 

Capture d’écran de 2025-02-26 10-33-47


Capture d’écran du 2025-02-26 10-37-10

src/components/inputs/reactHookForm/text/UniqueNameInput.tsx
```diff
@@ -158,7 +161,7 @@ export function UniqueNameInput({
             type="text"
             autoFocus={autoFocus}
             margin="dense"
-            fullWidth
```
basseche, line 161: you should not probably change that if sx argument is not null because it affects all Forms

Mathieu-Deharbe: Yep it would work, but I addded a props "fullWidth" defaulted to true. I find it clearer.

basseche (approved):

PR #706, opened 2025-02-28
Clean code

Tristan-WorkGH:
- `CancelButton` doesn't accept children
- add jsdoc
- remove unused types
- update directory-server API declaration
  - `ElementAttributes` have a lot of information that doesn't come from the server
  - the server accept a string, but the front recognize only `ElementType`


ayolab (approved):
Code : OK

PR #729, opened 2025-03-26
upgrade sonarscanner v3 -> v5

jonenst:


jonenst:
Sonar warnings are mostly just a reflect of the fact that everything is analyzed again. There are a few different analyzed issues but these are not the ones that are blocking.

PR #735, opened 2025-04-01
Set default value for currentName in UniqueNameInput component

ayolab:
Fix : to avoid triggering validation when no currentName is set


dbraquart (approved):
tests ok

PR #741, opened 2025-04-03
Jest doesn't support object `exports` in `package.json` in ESM mode

Tristan-WorkGH:
revert a change in #732


antoinebhs (approved):
Code OK

PR #749, opened 2025-04-11
Cleanup gitignore, use the same structure as other projects

jonenst:
This removes npm-debug.log* but we don't have it in other projects


sBouzols:
could you close this PR ? it's already done with #755

PR #751, opened 2025-04-14
add announcement banner

AbdelHedhili:


src/components/topBar/MessageBanner.tsx
Tristan-WorkGH, line 1: Why not using [Alter component](https://mui.com/material-ui/react-alert/) instead ? It have all we need.

src/components/announcement/AnnouncementBanner.tsx
```diff
@@ -0,0 +1,124 @@
...
+            if (id) {
+                setVisible(true);
+                if (duration) {
+                    // the previous timer is normally already cleared
+                    const bannerTimer = setTimeout(() => {
+                        setVisible(false);
+                    }, duration);
+                    return () => {
+                        clearTimeout(bannerTimer); // clear previous timer
+                    };
```
Meklo, line 92: Is this necessary to wrap it in an arrow function ?

Tristan-WorkGH: yes, this is the function to execute when the component unmount or the `useeffect` is trigger because one of the dependencies changed.  
https://18.react.dev/reference/react/useEffect

Meklo (approved):

PR #759, opened 2025-04-23
add material symbols icons

souissimai:


src/components/icons/LeftPanelCloseIcon.tsx
```diff
@@ -0,0 +1,23 @@
...
+    return (
+        <LeftPanelClose
+            style={{
+                width: 24,
+                height: 24,
+                fill: theme.icon?.color,
+            }}
+        />
+    );
+}
```
klesaulnier, line 23: I realised we don't need to create new theme property for this
"theme.palette.text.primary should" do the job

klesaulnier, line 23: I realised we don't need to create new theme property for this
"theme.palette.text.primary should" do the job

klesaulnier (approved):

PR #772, opened 2025-05-19
Add device hub icon

AAJELLAL:


src/components/icons/DeviceHubIcon.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { useTheme } from '@mui/material';
+
+interface DeviceHubProps {
+    style?: React.CSSProperties;
+}
+export function DeviceHubIcon({ style }: DeviceHubProps) {
+    const theme = useTheme();
+
+    const defaultStyle: React.CSSProperties = {
+        width: 15,
```
sBouzols, line 18: why 15 ? why not 24 has other icons ????

AAJELLAL: This is the default value I chose based on the Figma, you can always override it.
 For the other components, the value was set based on Figma with no option to override it.

souissimai (approved):

PR #775, opened 2025-05-21
Maintain loadflow parameters form actions buttons

Meklo:


dbraquart (approved):
tests ok on both Explore/srudy

PR #776, opened 2025-05-21
Extract user roles from token claims

achour94:


src/components/topBar/UserInformationDialog.tsx
```diff
@@ -46,6 +46,9 @@ function UserInformationDialog({ openDialog, user, onClose }: UserInformationDia
         }
     }, [openDialog, user]);
 
+    const rolesString = user?.profile?.profile ?? '';
+    const rolesList = rolesString ? rolesString.split('|').map((role) => role.trim()) : [];
```
Meklo, line 50:
```suggestion
    const rolesList = useMemo(() => {
        const rolesString = user?.profile?.profile ?? '';
        return rolesString ? rolesString.split('|').map((role) => role.trim()) : [];
    }, [user?.profile?.profile]);
```

achour94: No needed in our case, useMemo is more Useful for expensive recalculations  https://react.dev/reference/react/useMemo#skipping-expensive-recalculations

Meklo (approved):

PR #790, opened 2025-06-10
LOADFLOW type is now divided between two types, with and without ratio tap changers

klesaulnier:


src/components/parameters/common/computing-type.ts
```diff
@@ -7,6 +7,8 @@
 
 export enum ComputingType {
     LOAD_FLOW = 'LOAD_FLOW',
```
sBouzols, line 9: Why keeping `LOAD_FLOW` ComputingType ?

sBouzols: You should add some cases management in the later :
`formatComputingTypeLabel` function then no ?

PR #799, opened 2025-06-23
Add confirmation dialog for resetting load flow and voltage initialization parameters

achour94:


src/components/parameters/voltage-init/voltage-init-parameters-inline.tsx
```diff
@@ -76,14 +78,26 @@ export function VoltageInitParametersInLine({
...
-                messageTxt: error.message,
-                headerId: 'updateVoltageInitParametersError',
+        })
+            .catch((error) => {
+                snackError({
+                    messageTxt: error.message,
+                    headerId: 'updateVoltageInitParametersError',
+                });
+            })
+            .finally(() => {
```
sBouzols, line 88: Do you need to wait the end of the request to close the confirmation dialog ? I thnik you should move this outside of the finally

PR #811, opened 2025-07-23
Test with `data-testid`s

Tristan-WorkGH:


AmirHdm (approved):
Implemented as expected

PR #830, opened 2025-08-13
Add a test ID on Equipment Type Selector for Filter.

AmirHdm:


Tristan-WorkGH (approved):
`InputWithPopupConfirmation` --> `<Input {...props}` --> `<SelectInput {...props}` --> `<AutoComplete {...props}`

So seems OK

PR #835, opened 2025-08-19
Revert last modification on unique-name-validation useEffect

klesaulnier:
This change was breaking name validation in gridexplore when creating study from existing case
In this case, we need to make requests even if value.trim() === defaultValue


souissimai:
reverted code seems okay

PR #838, opened 2025-08-21
fix: handle config param api in commons-ui (app specific)

TheMaskedTurtle:


src/components/directoryItemSelector/utils.ts
```diff
@@ -30,7 +30,7 @@ export async function clearLastSelectedDirectory(): Promise<void> {
     localStorage.removeItem(LAST_SELECTED_DIRECTORY);
 
     try {
-        await updateConfigParameter(LAST_SELECTED_DIRECTORY, 'null');
+        await updateConfigParameter(COMMON_APP_NAME, LAST_SELECTED_DIRECTORY, 'null');
```
Tristan-WorkGH, line 33: This must be per application for this component. Did we even want to always memorize that "offline" in gridstudy, shared between studies?

Tristan-WorkGH (changes requested):

Tristan-WorkGH (approved):
OK

PR #847, opened 2025-09-09
Change limit reductions labels

carojeandat:
* Invert labels and tooltips in the limit reductions table
* Add tooltips on voltage ranges
* Remove excess English capitalization in the limit reductions table of the safety analysis settings

### DIFF
--- a/src/components/parameters/common/limitreductions/limit-reductions-table-form.tsx
+++ b/src/components/parameters/common/limitreductions/limit-reductions-table-form.tsx
@@ -15,31 +15,43 @@ import {
 } from './columns-definitions';
 import { CustomVoltageLevelTable } from '../voltage-level-table';
 
-const getLabelColumn = (limit: ITemporaryLimitReduction) => {
-    const lowBound = `${Math.trunc(limit.limitDuration.lowBound / 60)} min`;
-    const highBoundValue = Math.trunc(limit.limitDuration.highBound / 60);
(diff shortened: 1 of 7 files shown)


src/components/parameters/common/limitreductions/limit-reductions-table-form.tsx
```diff
@@ -15,31 +15,43 @@ import {
...
-    const lowerBoundClosed = limit.limitDuration.lowClosed ? '[' : ']';
-    const highBoundClosed = limit.limitDuration.highClosed || null ? ']' : '[';
-    return `${lowerBoundClosed}${lowBound}, ${highBound}${highBoundClosed}`;
-};
-
 export function LimitReductionsTableForm({ limits }: Readonly<{ limits: ILimitReductionsByVoltageLevel[] }>) {
     const intl = useIntl();
 
-    const getToolTipColumn = useCallback(
+    const getLabelColumn = useCallback(
```
sBouzols, line 21: We don't need to memoize those functions be cause it depends of any props or states in the `LimitReductionsTableForm` component. That's why `getLabelColumn` was in the global scope and not exported and I think we should move `getToolTipColumn` in global scope as well.
You need translation then make those functions return a <FormattedMessage/> React node type, it should work

antoinebhs (commented):
In addition to Sylvain's comments, one remark on translations

PR #863, opened 2025-09-19
Fix custom theme variables used

Tristan-WorkGH:
Following #855

Content:
* module-mui.d.ts → rework MUI theme augmentation to work inside commons-ui and in app including it
  - adapt Vite config to include dts in build
  - cleaning theme variable names across gridsuite code
  - include by default mui libs already in commons-ui dependencies
* Migrate some files of dome to typescript
* AgGrid theme option is optional because griddyna don't use aggrid
* AgGrid theme override endpoint are set to optional because each fronts don't set all and not the same
  - add condition to not inject incomplete style rules
  - remove redundant Box enclosure in `CustomAgGridTable`, already done by `CustomAGGrid`
* Follow MUI Guide to correctly set `CancelButton` custom color

### DIFF
--- a/src/components/inputs/reactHookForm/agGridTable/CustomAgGridTable.tsx
+++ b/src/components/inputs/reactHookForm/agGridTable/CustomAgGridTable.tsx
@@ -5,71 +5,73 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useFieldArray, useFormContext } from 'react-hook-form';
-import { Box, useTheme } from '@mui/material';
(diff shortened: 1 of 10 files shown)


src/components/inputs/reactHookForm/agGridTable/CustomAgGridTable.tsx
```diff
@@ -5,71 +5,75 @@
...
-        },
-        '& .ag-overlay-loading-center': {
-            border: 'none',
-            boxShadow: 'none',
-        },
-        '& .numeric-input': {
-            fontSize: 'calc(var(--ag-font-size) + 1px)',
-            paddingLeft: 'calc(var(--ag-cell-horizontal-padding) - 1px)',
-            width: '100%',
+const style = (customProps: CSSObject) =>
```
sBouzols, line 17: What is this `customProps` as parameter of the style object ?

PR #867, opened 2025-09-25
fix: rework typing and actually call broadcast on open

TheMaskedTurtle:
We were not actually calling the callback to broadcast the event on websocket opening.

To ba able to do that properly with TS I had to rework a bit the typing


sBouzols (approved):
Code Review OK

PR #924, opened 2025-11-13
Upgrade to MUI v6

FranckLecuyer:
## PR Summary

Upgrade to MUI v6 (but keeping the now deprecated Grid component)


src/components/grid/grid-item.tsx
```diff
@@ -14,8 +14,14 @@ export interface GridItemProps extends PropsWithChildren {
...
 export default function GridItem({ children, size = 6, alignItem = 'flex-start', tooltip }: Readonly<GridItemProps>) {
+    let newSize: number | boolean | 'auto' | undefined;
+    if (typeof size === 'number') {
+        newSize = size;
+    } else {
+        newSize = size === 'grow' ? true : size;
+    }
     return (
-        <Grid item xs={size} alignItems={alignItem}>
+        <Grid item xs={newSize} alignItems={alignItem}>
```
achour94: we can just use the right type of Grid v1, and it will be migrated into Grid2 in the future PR
import type { GridProps } from '@mui/material/Grid';
export interface GridItemProps extends PropsWithChildren {
    size?: GridProps['xs'];
    alignItem?: string;
    tooltip?: ReactNode;
}

achour94 (changes requested):

achour94 (approved):

antoinebhs (approved):

PR #1096, opened 2026-04-14
Chore(auth): migrate from oidc-client to oidc-client-ts

achour94:


src/components/authentication/utils/authService.ts
```diff
@@ -38,7 +38,7 @@ type CustomUserManager = UserManager & {
 
 declare global {
     interface Window {
-        OIDCLog?: Log;
+        OIDCLog?: typeof Log;
     }
 }
```
TheMaskedTurtle: Actually I don't think this is good to keep that in production code

package-lock.json
TheMaskedTurtle, line 1: Your node/npm version must not be up-to-date, you shouldn't have so muche changes in package-lock

TheMaskedTurtle (changes requested):

TheMaskedTurtle (approved):

PR #1104, opened 2026-04-20
Set clear rules for composite selection behaviour

Meklo:

### DIFF
--- a/src/components/network-modification-table/renderers/select-cell.tsx
+++ b/src/components/network-modification-table/renderers/select-cell.tsx
@@ -5,7 +5,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import React, { useCallback, useMemo } from 'react';
+import React, { useCallback } from 'react';
 import { Checkbox } from '@mui/material';
 import { Row, Table } from '@tanstack/react-table';
 import { networkModificationTableStyles } from '../network-modification-table-styles';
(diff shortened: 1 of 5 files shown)


src/components/network-modification-table/renderers/select-cell.tsx
```diff
@@ -16,66 +16,48 @@ interface SelectCellRendererProps {
...
-                    delete nextSelection[row.id];
-                } else {
-                    // was unselected, now toggled on
-                    nextSelection[row.id] = true;
-                }
+                row.toggleSelected(targetSelected);
             }
 
             if (meta) {
                 meta.lastClickedIndex.current = currentIndex;
```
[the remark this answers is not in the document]
Mathieu-Deharbe: @Meklo I am afraid that codderabbit is right. Look : 

Image

src/components/network-modification-table/use-modifications-selection.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    const visit = (node: ComposedModificationMetadata): boolean => {
+        const children = node.subModifications;
+        // Exception for composite having only one children we want to be able to select the modification individually
+        // without selecting the composite
+        if (!children || children.length === 0 || children.length === 1) {
+            return next[node.uuid];
+        }
+        // Recurse first so nested composites are resolved bottom-up.
+        const everyChildSelected = children.map((child) => visit(child)).every(Boolean);
```
Mathieu-Deharbe: Doesn't this do the same ? : 
```suggestion
        const everyChildSelected = children.every((child) => visit(child));
```

Mathieu-Deharbe (commented):
I prefer like it is now : when the children are all selected, the composite parent is not : you have to select it anually specifically :

Image

But the fact that this is deselected and not isIndeterminate is a bit confusing. The PO will decide.

src/components/network-modification-table/use-modifications-selection.ts
Meklo: It does, I rolled it back to please sonar

Mathieu-Deharbe: I think that sonar was displeased because of the useless map of your first version which was : 

`const everyChildSelected = children.map((child) => visit(child)).every((child) => child);`

Mine merges the two, I think it will be fine for sonar. But who knows...

Mathieu-Deharbe (approved):

src/components/network-modification-table/use-modifications-selection.ts
Meklo: Let's try it out, I prefered the solution without Boolean too

Meklo: It seems to please it :)

PR #1108, opened 2026-04-23
Add business error code for monitor

FranckLecuyer:


src/translations/en/businessErrorsEn.ts
```diff
@@ -92,4 +92,6 @@ export const businessErrorsEn = {
     'diagram.noVoltageLevelFound': 'No voltage level found for this network area diagram',
     'dynamicMarginCalculation.providerNotFound': 'Dynamic margin calculation provider not found.',
     'dynamicMarginCalculation.loadFilterNotFound': 'Some load filters do not exist: {filterUuids}',
+
+    'monitor.server.differentProcessConfigType': 'Cannot compare 2 different process config types',
```
antoinebhs: I think we need to use processConfigEntity1Type and processConfigEntity2Type? Like         'You need to reduce the number of voltage levels to be displayed in the network area diagram (current {nbVoltageLevels}, maximum {maxVoltageLevels})', ?
Because we should have it in the buisness error?

FranckLecuyer: :+1: 
My bad

antoinebhs (approved):

PR #1118, opened 2026-04-30
Fix GridSuiteModule "app" type

klesaulnier:


antoinebhs (approved):
linked to https://github.com/gridsuite/study-server/blob/main/src/main/java/org/gridsuite/study/server/dto/AboutInfo.java#L22

PR #1119, opened 2026-05-04
fix validation message localization in production

khouadrired:
In production bundles, the order of module execution can vary. We discovered that several Yup schemas were defined as top-level constants. This caused them to be initialized the moment their file was loaded—often before the main application had a chance to call `yup.setLocale`.
Because Yup "bakes" its error messages at the moment of schema creation, these schemas were stuck with default English strings, ignoring the subsequent localization configuration.

### DIFF
--- a/src/components/parameters/dynamic-margin-calculation/loads-variations-parameters.tsx
+++ b/src/components/parameters/dynamic-margin-calculation/loads-variations-parameters.tsx
@@ -31,27 +31,28 @@ import { NAME } from '../../inputs';
 import { ParameterTableField } from '../common/parameter-table-field';
 import { DndColumn, DndColumnType } from '../../dnd-table-v2';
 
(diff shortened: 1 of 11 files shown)


src/components/inputs/reactHookForm/utils/functions.tsx
```diff
@@ -14,7 +14,7 @@ export function genHelperError(...errors: any[]) {
     if (inError) {
         return {
             error: true,
-            helperText: <FormattedMessage id={inError} />,
+            helperText: <FormattedMessage id="Required" defaultMessage="Required" />,
```
antoinebhs: @khouadrired this wll change all the errors in the app: is that what we want? confirm with PO?

khouadrired: yes , that's what we want to be shown when a required field is not filled.

PR #1121, opened 2026-05-04
Simplify NetworkModificationsTable column API

flomillot:
## Summary

Accept a stable `columns` array prop instead of a `createAllColumns` factory called inside an internal `useMemo`, so the parent owns when columns are rebuilt.

- Memoize `nameHeaderProps` from the individual fields and forward it through react-table `meta` along with `isRowDragDisabled` and `modificationsCount` so column renderers can read them via `table.options.meta`.
- Augment `TableMeta` with the new fields and `ColumnMeta` with `cellStyle`.

Paired with https://github.com/gridsuite/gridstudy-app/pull/3931, which adopts this API to stop unmounting cells on every columns rebuild.


Meklo (approved):
Small linter issue to fix

PR #1122, opened 2026-05-05
rename modification

souissimai:


Mathieu-Deharbe (changes requested):
When I validate, the text stay displayed briefly like it used to be before the editing and then the new texte replaces it : 

Image

src/components/network-modification-table/renderers/name-cell.tsx
```diff
@@ -85,16 +177,60 @@ export function NameCell({ row }: Readonly<NameCellProps>) {
...
-                            sx={mergeSx(
-                                networkModificationTableStyles.modificationLabel,
-                                createModificationNameCellStyle(row.original.activated)
-                            )}
+                            sx={{
+                                display: 'inline-flex',
+                                width: frozenWidthPx ? `${frozenWidthPx}px` : `${MIN_CHAR_WIDTH}ch`,
+                                maxWidth: '100%',
+                                flexShrink: 0,
+                            }}
```
Mathieu-Deharbe: I would add this to keep the same text appearance between the regular modification name  text and the edited modification text : 

```suggestion
                            sx={mergeSx(networkModificationTableStyles.modificationLabel, {
                                display: 'inline-flex',
                                width: frozenWidthPx ? `${frozenWidthPx}px` : `${MIN_CHAR_WIDTH}ch`,
                                maxWidth: '100%',
                                flexShrink: 0,
                            })}
```

But the only important thing is to have  `paddingLeft: '0.5vw',`. Otherwise when you begin to edit the texte slide lightly to the left and then when you validate, it slides again to the right.

src/components/network-modification-table/renderers/name-cell.tsx
```diff
@@ -45,14 +59,92 @@ export function NameCell({ row }: Readonly<NameCellProps>) {
...
+
+    const handleKeyDown = useCallback(
+        (e: React.KeyboardEvent<HTMLInputElement>) => {
+            if (e.key === 'Enter') {
+                const trimmed = draftName.trim();
+                if (trimmed !== '' && trimmed !== compositeName) {
+                    onEditNameCell?.(row.original, trimmed);
+                }
+                stopEditing();
+                inputRef.current?.blur();
```
Mathieu-Deharbe, line 135: This looks identical to what is inside `handleBlur`. Shouldn't this be inside a shared function ?

src/components/network-modification-table/renderers/name-cell.tsx
```diff
@@ -45,14 +59,92 @@ export function NameCell({ row }: Readonly<NameCellProps>) {
...
+        try {
+            return (JSON.parse(row.original.messageValues)?.name as string) ?? '';
+        } catch {
+            return '';
+        }
+    }, [isComposite, row.original.messageValues]);
+
+    const [isEditing, setIsEditing] = useState(false);
+    const [draftName, setDraftName] = useState('');
+    const [frozenWidthPx, setFrozenWidthPx] = useState<number | null>(null);
```
Mathieu-Deharbe: I think this would be better to name this after the fact that this is an edit box. Maybe ? : 
```suggestion
    const [inputBaseWidthPx, setInputBaseWidthPx] = useState<number | null>(null);
```

src/components/network-modification-table/renderers/name-cell.tsx
```diff
@@ -85,16 +177,60 @@ export function NameCell({ row }: Readonly<NameCellProps>) {
...
+                                    },
+                                }}
+                            />
                         </Box>
-                    </CustomTooltip>
+                    ) : (
+                        /* Read mode */
+                        <CustomTooltip disableFocusListener disableTouchListener title={label}>
+                            <Box
+                                ref={isComposite ? labelRef : undefined}
```
Mathieu-Deharbe: `undefined` when `isComposite` because it is not useful when the name-cell doesn't belong to a composite ? 

It is logical but also a bit strange to have a ref pointing to null despite having something to ref. It makes me think that composites and not composites are now very different and shouldn't share everything.

src/components/network-modification-table/renderers/name-cell.tsx
```diff
@@ -45,14 +59,92 @@ export function NameCell({ row }: Readonly<NameCellProps>) {
...
+
+    const handleLabelClick = useCallback(
+        (e: React.MouseEvent) => {
+            e.stopPropagation();
+
+            // Init draft with current saved name
+            setDraftName(compositeName);
+            // Freeze input width at open time
+            if (labelRef.current) {
+                const style = window.getComputedStyle(labelRef.current);
```
Mathieu-Deharbe: Sonar asks you to use `globalThis` instead of `window`, but I prefer your version so do as you wish.

souissimai: ok

src/components/network-modification-table/renderers/name-cell.tsx
souissimai: refactored

src/components/network-modification-table/renderers/name-cell.tsx
souissimai: yes

src/components/network-modification-table/renderers/name-cell.tsx
souissimai: yes

Mathieu-Deharbe (approved):

PR #1123, opened 2026-05-06
chore: move battery creation form to commons UI

dbraquart:


src/components/inputs/reactHookForm/booleans/RadioInput.tsx
```diff
@@ -26,7 +26,7 @@ export interface RadioInputProps {
 export function RadioInput({ name, label, id, options, formProps }: Readonly<RadioInputProps>) {
     const {
         field: { onChange, value },
-    } = useController({ name });
+    } = useController({ name, defaultValue: 'MINMAX' });
```
TheMaskedTurtle: This is a common component, it feels clumsy. Do we really have to put it here ?

TheMaskedTurtle (approved):

src/components/inputs/reactHookForm/booleans/RadioInput.tsx
dbraquart: good point. at the beginning I put empty string, and Rabbit suggested MINMAX, which was a wrong advice !

TheMaskedTurtle (approved):

PR #1124, opened 2026-05-07
refactor(voltage-level-creation): remove `CreateSwitchesDialog` and replace it with a simplified inline form (#1101)

TheMaskedTurtle:


dbraquart (commented):
code: ok
tests: ok

open question before approve: could we have a max value for section number ? I tried 1000 and my browser was in total panic !

dbraquart (approved):
code & tests: ok

PR #1132, opened 2026-05-13
Move network-modification-table renderers into commons-ui

flomillot:
## Summary

Brings the per-row cells used by `NetworkModificationsTable` (description, switch, root-network chip and the `cell-renderers` adapter) into commons-ui alongside the existing renderers.

- Cells receive `studyUuid`, `currentNodeId`, `isDisabled` and the root-network metadata via the table's `TableMeta` rather than reading them from the gridstudy Redux store, so the lib stays state-agnostic.
- `ColumnMeta` carries only per-column values — `cellStyle` and a `onChange` callback; everything shared across columns (`rootNetworks` list, `currentRootNetworkUuid`, `modificationsToExclude` / `setModificationsToExclude`) lives in `TableMeta`.
- Ports the `setModificationMetadata` and `updateModificationStatusByRootNetwork` services from gridstudy-app to commons-ui.
- Adds `updateNetworkModificationsMetadata`, a service that updates network-modification metadata (e.g. the name) directly on the network-modification-server, with no study/node context — used by GridExplore to rename composite modifications nested inside composite modifications.
- `NameCell` renames composite modifications inline: it receives an `onChange` callback through its column's `ColumnMeta` (the consumer provides the actual persistence) and keeps an optimistic local name, re-synced from the server data with a rollback on failure.
- Adds a minimal `RootNetworkRowInfo` type so consumers can pass their own richer root-network objects (structurally compatible).

Companion PR on gridstudy-app: https://github.com/gridsuite/gridstudy-app/pulls?q=is%3Apr+head%3Aperf%2Fmove-network-modification-renderers

## Post-move fixes (applied by CodeRabbit autofix on pre-existing code)

Several issues in the moved renderers and service were caught and fixed during the migration :

- **MUI deep import** (`description-cell.tsx`): replaced `@mui/material/IconButton` deep import with the root `@mui/material` import to comply with the `no-restricted-imports` ESLint rule enforced in this repo.
- **Function declaration style** (`description-cell.tsx`, `root-network-chip-cell.tsx`, `switch-cell.tsx`): converted arrow-function component exports (`export const Foo: FunctionComponent = () => {}`) to function declarations (`export function Foo(...) {}`) required by the `react/function-component-definition` Airbnb ESLint rule.
- **Impure state updater** (`root-network-chip-cell.tsx`): moved the `updateModificationStatusByRootNetwork` backend call out of the `setModificationsToExclude` updater callback (which must stay pure), computing the next state first, then applying the optimistic update and rolling back in `catch` on failure.
- **`undefined` guard in `setModificationMetadata`** (`networkModification.ts`): added an early rejection when `modificationUuid` is absent to prevent sending `uuids=undefined` to the backend.


flomillot:
@coderabbitai resolve the comments you fixed. Also update the PR description by adding some lines about your fix considering that it was on pre existing code that was just moved.

src/components/network-modification-table/renderers/name-cell.tsx
```diff
@@ -83,17 +99,31 @@ export function NameCell({ row, onEditNameCell }: Readonly<NameCellProps>) {
...
+            // Optimistic update: adopt the new name immediately (the label derives from
+            // it); persistence is delegated to the consumer via onChange, and we
+            // roll back to the server value if it fails.
+            setCompositeName(newName);
+            Promise.resolve(onChange?.(row.original, newName)).catch((error) => {
+                setCompositeName(savedCompositeName); // rollback
+                snackWithFallback(snackError, error, { headerId: 'networkModificationRenamingError' });
+            });
+        },
+        [onChange, row.original, savedCompositeName, snackError]
```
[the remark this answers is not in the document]
flomillot: too low probability to happen

src/components/network-modification-table/renderers/cell-renderers.tsx
```diff
@@ -0,0 +1,130 @@
...
+export function RootNetworkHeaderRenderer({ column, table }: HCtx) {
+    const { meta } = table.options;
+    // `column.id` is the rootNetworkUuid (set in createRootNetworksColumns).
+    const isCurrentRootNetwork = !!meta?.currentRootNetworkUuid && column.id === meta.currentRootNetworkUuid;
+    if (!isCurrentRootNetwork || (meta?.modificationsCount ?? 0) < 1) {
+        return null;
+    }
+    const currentRootNetworkTag = meta?.rootNetworks?.find(
+        (r) => r.rootNetworkUuid === meta.currentRootNetworkUuid
+    )?.tag;
```
Meklo: currentRootNetworkTag is now recomputed on every renders while before it was passed down as metadata, it's minor but could it be computed outside the renderer ?

src/components/network-modification-table/renderers/cell-renderers.tsx
```diff
@@ -0,0 +1,130 @@
...
+ * factories called inside a hook — defining them inline there would produce fresh references.
+ *
+ * Dynamic values are routed via react-table's `meta`: table-wide via `table.options.meta`,
+ * per-column via `column.columnDef.meta`.
+ */
+
+type CCtx = CellContext<ComposedModificationMetadata, unknown>;
+type HCtx = HeaderContext<ComposedModificationMetadata, unknown>;
+
+export function DragHandleRenderer({ table }: CCtx) {
```
Meklo, line 36: Sonar issues on a lot of declarations, encapsulating CCtx and HCtx uses as well as DescriptionCell, RootNetworkChipCell and SwitchCell props in Readonly should please it

src/components/network-modification-table/renderers/cell-renderers.tsx
flomillot: Actually it was already the case. The columns were recreated every time.
But I agree now it's done for every cell. Which is not great. However, it's not really a computing and it's a very quick find, which is imperceptible for the user. 
Also we already add a similar line here. 
https://github.com/gridsuite/commons-ui/blob/6fcb8435dcb4c8803dd42968e7ff6dc34397b246/src/components/network-modification-table/renderers/cell-renderers.tsx#L114
If you want, I can create a use memo, but I'm not sure it's a big win. 
I don't want to compute it outside because that will mean we will have duplicate information in the props because root networks already contain everything.

Meklo (approved):

src/components/network-modification-table/renderers/cell-renderers.tsx
flomillot: Actually in discussion to remove this rule.

PR #1133, opened 2026-05-15
feat(CustomAgGridTable): expose useFieldArray via ref and relax props

flomillot:
## Summary

- Wrap `CustomAgGridTable` in `forwardRef` + `useImperativeHandle` so a parent can drive the internal `useFieldArray` (`replace`, `append`, ...) without declaring a second `useFieldArray` on the same name (forbidden by react-hook-form).
- Relax the props type so only `columnDefs` is required. `defaultColDef`, `pagination`, `paginationPageSize`, `alwaysShowVerticalScroll`, `stopEditingWhenCellsLoseFocus`, `rowSelection`, `overrideLocales` are now optional. `loading` is exposed too.
- Default `stopEditingWhenCellsLoseFocus` to `true` (matches every current caller).
- Remove hardcoded `rowDragManaged` (not compatible with pagination per AG Grid docs) and `rowDragEntireRow`. Consumers opt into row drag via `rowDrag: true` on a column.
- Drop the now-redundant `stopEditingWhenCellsLoseFocus` prop on `ExplicitNamingFilterForm`.

Motivated by gridstudy-app's tabular-modification dialog that needs to load CSV-parsed rows imperatively into the grid. Consumer PRs:
- gridsuite/gridstudy-app#3948
- gridsuite/gridexplore-app#867


src/components/composite/agGridTable/CustomAgGridTable.tsx
```diff
@@ -71,176 +71,175 @@ const style = (customProps: any) => ({
...
-    return (
-        <>
-            <Box className={theme.aggrid.theme} sx={style(cssProps).grid}>
-                <CustomAGGrid
-                    rowData={rowData}
-                    onGridReady={onGridReady}
-                    cacheOverflowSize={10}
-                    rowSelection={rowSelection ?? 'multiple'}
-                    rowDragEntireRow
-                    rowDragManaged
```
ghazwarhili, line 218: why rowDragManaged is removed ?

src/components/composite/agGridTable/CustomAgGridTable.tsx
```diff
@@ -71,176 +71,175 @@ const style = (customProps: any) => ({
...
+
+        return (
+            <>
+                <Box className={theme.aggrid.theme} sx={style(cssProps).grid}>
+                    <CustomAGGrid
+                        rowData={rowData}
+                        onGridReady={onGridReady}
+                        cacheOverflowSize={10}
+                        rowSelection={rowSelection ?? 'multiple'}
+                        onRowDragMove={(e) => move(getIndex(e.node.data), e.overIndex)}
```
ghazwarhili, line 216: onRowDragEnd is removed and replaced by onRowDragMove ?
This changes the behavior significantly: move() is now called on every drag frame instead of once on drop => this triggers multiple RHF state updates and re-renders during the drag !

src/components/composite/agGridTable/CustomAgGridTable.tsx
```diff
@@ -71,176 +71,175 @@ const style = (customProps: any) => ({
...
-    const onRowDataUpdated = () => {
-        setNewRowAdded(false);
-        if (gridApi?.api) {
-            // update due to new appended row, let's scroll
-            const lastIndex = rowData.length - 1;
-            gridApi.api.paginationGoToLastPage();
-            gridApi.api.ensureIndexVisible(lastIndex, 'bottom');
-        }
-    };
+        );
```
ghazwarhili, line 122: Prettier formatting to be fixed

src/components/composite/agGridTable/CustomAgGridTable.tsx
flomillot: As said in the summary it's not compatible with pagination according to the documentation. It was working apparently but not sure it was working very well.

src/components/composite/agGridTable/CustomAgGridTable.tsx
flomillot: It was done internally by AGGrid and now we just do it manually. 
This is how to do it when you don't use rowDragManaged. At least this is how they do in the documentation. 
Did you notice a difference in terms of performance ? On my computer it's fine, but maybe not on everyone's

src/components/composite/agGridTable/CustomAgGridTable.tsx
flomillot: Ah yes, my auto formatting is broke, done.

ghazwarhili (approved):

PR #1152, opened 2026-05-27
add isAuthenticated to fix no user token available

ghazwarhili:
## PR Summary
 If the provider mounted before OIDC authentication had produced a token, it logged "Skipping Notification WebSockets: no user token available" and never retried once the token became available leaving the user with no notifications until a full page reload.
 
 Fix: 
 Expose auth state reactively from commonStore: add getUser() and subscribeToUserState(), plus a subscribe() method on the CommonStore interface.In NotificationsProvider, derive isAuthenticated with useSyncExternalStore(subscribeToUserState, () => getUser() !== null), include it in the effect guard (!isAuthenticated || !token) and in the dependency array.

The effect now re-runs on any authentication change: WebSockets are opened on login and torn down on logout, instead of being established only once at mount.


src/features/notifications/NotificationsProvider.tsx
```diff
@@ -37,17 +37,18 @@ export function NotificationsProvider({ urls, children }: PropsWithChildren<Noti
         removeListener: removeListenerOnReopen,
     } = useListenerManager<ListenerOnReopen>(urls);
 
+    const isAuthenticated = useSyncExternalStore(subscribeToUserState, () => getUser() !== null);
```
flomillot, line 45: Why don't you just subscribe to the token and remove the get user token below Inside the use effect ? 
I think we fixed the unexpected refresh every minute, If I remember well ?

src/redux/commonStore.ts
```diff
@@ -26,6 +27,13 @@ export function setCommonStore(store: CommonStore): void {
     commonStore = store;
 }
 
+export function getUser(): User | null {
```
flomillot, line 30: Do you really need it ? You can't just use get user token and check if it's undefined or null ?

src/features/notifications/NotificationsProvider.tsx
ghazwarhili: getUser() !== null stays stable across renew (oidc fires addUserLoaded with the new user, it doesn't null it out), so it only flips on real login/logout. That's what gives us 'create the WS once when auth is ready' without recreating it on renew.

So subscribing on the token instead of the user isn't equivalent here; it would bring back the refetch-on-renew

src/redux/commonStore.ts
ghazwarhili: yup

src/features/notifications/NotificationsProvider.tsx
flomillot: Okay, after discussion the refresh is not fixed anyway.

src/redux/commonStore.ts
flomillot: After discussion it's because the token is set to null at every renew.

flomillot (approved):

flomillot:
Tests OK

src/features/notifications/NotificationsProvider.tsx
```diff
@@ -37,9 +37,10 @@ export function NotificationsProvider({ urls, children }: PropsWithChildren<Noti
...
     useEffect(() => {
         const token = getUserToken();
-        if (!token) {
+        if (!isAuthenticated || !token) {
```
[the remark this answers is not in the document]
ghazwarhili: The premise here isn't right,  the token isn't baked into the URL at WS creation. 
We're passing a function to ReconnectingWebSocket:

  new ReconnectingWebSocket(() => appendTokenToUrl(url, getUserToken() ?? ''), [], { ... })

jonenst (approved):

PR #1154, opened 2026-05-28
chore(yup): set up good practices for yup module usage

TheMaskedTurtle:
## PR Summary

Main points:
- New barrel to export configureYup() method to ba called in each appliation before schema creation
- Use everywhere default yup import and not re-exported internal module
- Refactor imports to not import the whole yup module everytime  
- Remove hard-coded validation message keys when they will be setup by configureYup method (so mainly for required fields and type errors)


achour94 (approved):
don't forget to update the version in package.json

PR #1166, opened 2026-06-04
refactor(csv): replace CsvUploader with inline CsvPicker and move CSV buttons into the table

flomillot:
## Context

Reworks the shared CSV import/export building blocks used by gridstudy-app and
gridexplore-app. The previous modal `CsvUploader` is replaced by an inline
`CsvPicker`, and the CSV template/download buttons are moved into the table's
bottom bar so every `CustomAgGridTable` gets a consistent CSV experience.

### New `CsvPicker` (`src/components/ui/csvPicker/`)
- Inline file picker built on `react-papaparse`, replacing the modal `CsvUploader`.
- Generic over the parsed row type (`CsvPicker<TData>`).
- Default parse config: `header: true`, `skipEmptyLines: true`, `comments: '#'`,
  delimiter derived from the language via the shared `getCsvDelimiter` helper
  (`;` in FR, `,` otherwise); overridable via `parseConfig`.
- Validations: empty file (`noDataInCsvFile`), header mismatch
  (`wrongCsvHeadersError`, order-independent via `equalsArrayAnyOrder`), and optional
  `maxLineNumber` (`tooManyLinesInCsvFile`).
- Import flow: `onComplete` (replace) or `onAppend`/`onReplace`. In the latter mode,
  `CsvPicker` inspects the current rows returned by the `getTableData` prop (via
  `hasNonEmptyRows`) and, when the table is non-empty, opens a keep/replace confirmation
  dialog (`CsvPickerConfirmationDialog`). The empty-rows check lives inside `CsvPicker`,
  so consumers only provide `getTableData` instead of duplicating a predicate.
- `CsvPickerProps` callbacks are typed as a mutually-exclusive union: either
  `onComplete` alone, or `onAppend` + `onReplace` together — the two modes cannot be
  mixed and at least one must be provided. `onFileChange` is also fired from the
  confirmation dialog's replace/append actions so the selected file name stays in sync.

### Table CSV buttons (`BottomTableButtons`, replaces `BottomRightButtons`)
- Renders the CSV template (`getTemplateData`) and download (`getTableData`) buttons
  from `csvProps`; `DownloadCSV` is disabled while the table is empty (`hasTableData`).
- New `CsvProps` type in `agGridTable-utils.ts`. The `hasNonEmptyRows` helper lives in
  `utils/functions.ts` so it can be shared with `ui` components (which must not import from `composite`).

### `CustomAgGridTable`
- Selection column pinned left with the row-drag handle (`selectionColumnDef`).
- Selection checkboxes greyed (theme `text.secondary`) instead of the AG Grid accent;
  removed the old `--ag-*` colour overrides.
- `hasTableData` computed internally and injected into `csvProps`.
- `csvProps` and `cssProps` made optional.
- Grid container relies on `height: 100%` (dropped the redundant `width: 'auto'`) so the
  table fills its flex parent when no `cssProps` height is provided.
- `rowSelection` defaults to the AG Grid object form (`{ mode: 'multiRow' }`).

### `ExplicitNamingFilterForm`
- Migrated to `CsvPicker` with the keep/replace confirmation
  (`onAppend`/`onReplace`/`getTableData`).
- Flex-fill column layout; error `Alert` in its own grid; `getTemplateData`/`getTableData`
  for template + export.
- Equipment type selector width capped (`width: 400, maxWidth: '100%'`).

### Misc
- New `getCsvDelimiter(language)` helper in `utils/langs.ts`, reused everywhere the
  language-based `;`/`,` separator was previously inlined (and shared with the apps).
- `equalsArrayAnyOrder` now accepts a `null` second argument.
- CsvPicker error messages (no data / wrong headers / too many lines) now include the imported
  file name (quoted with `"` in French, `'` in English).
- `CsvPicker` reports the file via `onFileChange` only on a successful import; on a parse error
  the previously selected file name is kept (the table data is not replaced either).
- `numericEditor.ts` → `numericEditor.tsx` (JSX cell editor).
- Translations: `csv*` add `DownloadCSV`/`tooManyLinesInCsvFile`, reword `UploadCSV`,
  drop `ImportCSV`; `commonButton*` add `add`/`replace`.
- Drop the `react-csv-downloader` dependency (now using `react-papaparse`'s downloader).


src/components/composite/agGridTable/cellEditors/numericEditor.tsx
```diff
@@ -0,0 +1,92 @@
...
+ * Numeric keys still start editing; keys pressed while already editing are left to the input's own onKeyDown handling.
+ */
+export const suppressNonNumericKeyboardEvent = (params: SuppressKeyboardEventParams) =>
+    !params.editing && params.event.key.length === 1 && !isCharNumeric(params.event.key);
+
+// commit the value as a number, accepting "," as a decimal separator (null when empty/invalid)
+const toNumber = (text: string) => {
+    const result = parseFloat(text.replace(',', '.'));
+    return Number.isNaN(result) ? null : result;
+};
```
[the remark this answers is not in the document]
flomillot: It was the previous behavior and I let it like that.

src/components/composite/filter/explicitNaming/ExplicitNamingFilterForm.tsx
```diff
@@ -208,63 +228,83 @@ export function ExplicitNamingFilterForm({
...
+                        label="UploadCSV"
+                        header={csvFileHeaders}
+                        language={language ?? LANG_SYSTEM}
+                        selectedFile={selectedFile}
+                        onFileChange={setSelectedFile}
+                        onFileError={setSelectedFileError}
+                        hasExistingData={hasExistingData}
+                        onAppend={(results) => tableRef.current?.append(getDataFromCsvFile(results.data))}
+                        onReplace={(results) => tableRef.current?.replace(getDataFromCsvFile(results.data))}
                     />
```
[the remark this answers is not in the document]
flomillot: I will fix it using the disabled prop.

src/components/ui/csvPicker/csv-picker.tsx
```diff
@@ -0,0 +1,128 @@
...
+                onFileError(undefined);
+                if (onAppend && onReplace) {
+                    if (hasExistingData?.()) {
+                        setPendingImport({ results, file: acceptedFile });
+                    } else {
+                        onReplace(results, acceptedFile);
+                    }
+                } else {
+                    onComplete?.(results, acceptedFile);
+                }
```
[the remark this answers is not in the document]
flomillot: Yes, I know it is on pure pose.

src/components/composite/agGridTable/cellEditors/numericEditor.tsx
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * NB: the `eventKey`-driven logic below (seeding the initial value from the triggering key in the
+ * useState/useEffect) only matters when editing is started by typing a character on a focused cell,
+ * i.e. when the column uses `singleClickEdit: false`. With `singleClickEdit: true` (what we mostly
+ * use), editing starts on click, `eventKey` is null, and those branches are effectively inert.
+ *
+ * Columns using this editor should also set `suppressKeyboardEvent: suppressNonNumericKeyboardEvent`
+ * so that a non-numeric key on a focused cell doesn't start an edit.
+ */
+export function NumericEditor({ value, onValueChange, eventKey }: CustomCellEditorProps) {
```
basseche, line 41:
```suggestion
export function NumericEditor({ value, onValueChange, eventKey }: Readonly<CustomCellEditorProps>) {
```

Add Readonly before props type

src/components/composite/agGridTable/agGridTable-utils.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) 2024, RTE (http://www.rte-france.com)
```
basseche: 2026 ?

src/components/ui/csvPicker/csv-picker.tsx
```diff
@@ -0,0 +1,128 @@
...
+    language,
+    parseConfig,
+    selectedFile,
+    onFileChange,
+    onFileError,
+    onComplete,
+    onAppend,
+    onReplace,
+    hasExistingData,
+}: CsvPickerProps<TData>) {
```
basseche, line 55: mark prop read only

src/utils/algos.ts
```diff
@@ -31,3 +31,15 @@ export function equalsArray<T>(a: T[] | null, b: T[]) {
...
 }
+
+export function equalsArrayAnyOrder<T>(a: T[] | null, b: T[]) {
+    if (a === b) {
+        return true;
+    }
+    if (!a || !b || a.length !== b.length) {
+        return false;
+    }
+    const sortedA = [...a].sort();
```
basseche, line 42: did you check those warnings ?

basseche (changes requested):
You should check out sonar warnings et resolve them

src/components/composite/agGridTable/cellEditors/numericEditor.tsx
flomillot: As discussed we shouldn't do it

src/utils/algos.ts
flomillot: The point is to compare the array, whatever the sort is. So I don't see the point of implementing a compare method.

basseche (approved):

src/components/composite/agGridTable/cellEditors/numericEditor.tsx
```diff
@@ -0,0 +1,92 @@
...
+ * Copyright (c) 2026, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+import { type KeyboardEvent, useEffect, useRef, useState } from 'react';
+import { type CustomCellEditorProps } from 'ag-grid-react';
+import { KeyCode, type SuppressKeyboardEventParams } from 'ag-grid-community';
+
+const isCharNumeric = (charStr: string | null | undefined) => !!charStr && /\d|,|\./.test(charStr);
```
ghazwarhili, line 11: the negative number ( - ) is not included in isCharNumeric ?

src/components/composite/agGridTable/cellEditors/numericEditor.tsx
```diff
@@ -0,0 +1,92 @@
...
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+
+    const handleChange = (event: { target: HTMLInputElement }) => {
+        setText(event.target.value);
+        onValueChange(toNumber(event.target.value));
+    };
+
+    const handleKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
+        // block single non-numeric characters, let everything else through (navigation, backspace…)
```
ghazwarhili, line 70: This would allow us to skip validation for non-character keys and only apply the numeric check to actual character input 
if (event.key.length !== 1) {
return; 
    }

src/components/composite/agGridTable/cellEditors/numericEditor.tsx
```diff
@@ -0,0 +1,92 @@
...
+        if (event.key.length === 1 && !isCharNumeric(event.key)) {
```
ghazwarhili, line 71: We could handle the negative number here and allow - in the first position
const isLeadingMinus = event.key === '-' && input.selectionStart === 0 && !input.value.includes('-');

src/utils/algos.ts
```diff
@@ -31,3 +31,15 @@ export function equalsArray<T>(a: T[] | null, b: T[]) {
     }
     return true;
 }
+
+export function equalsArrayAnyOrder<T>(a: T[] | null, b: T[]) {
```
ghazwarhili: b: T[] | null

src/components/composite/agGridTable/CustomAgGridTable.tsx
ghazwarhili: rowSelection={rowSelection ?? { mode: 'multiRow' }}

ghazwarhili: onSelectionChanged={(event) => { setSelectedRows(event.api.getSelectedRows())

src/components/composite/agGridTable/BottomTableButtons.tsx
```diff
@@ -0,0 +1,132 @@
...
+}: BottomTableButtonsProps) {
+    return (
+        <>
+            <Grid container paddingTop={1} alignItems="center" spacing={1}>
+                {csvProps?.getTemplateData && (
+                    <Grid>
+                        <CsvDownloadButton
+                            data={csvProps.getTemplateData}
+                            fileName={csvProps.fileName}
+                            delimiter={csvProps.language === LANG_FRENCH ? ';' : ','}
```
ghazwarhili: Maybe we could extract the delimiter into a constant to avoid repeating the language check
const delimiter = csvProps.language === LANG_FRENCH ? ';' : ',';

src/components/ui/csvPicker/csv-picker.tsx
```diff
@@ -0,0 +1,135 @@
...
+    );
+
+    return (
+        <>
+            <CSVReader
+                config={{
+                    header: true,
+                    skipEmptyLines: true,
+                    comments: '#',
+                    delimiter: language === LANG_FRENCH ? ';' : ',',
```
ghazwarhili: maybe we can add transformHeader: (h: string) => h.replace(/^\uFEFF/, '').trim(), It removes unwanted whitespace in headers

src/components/composite/agGridTable/cellEditors/numericEditor.tsx
```diff
@@ -0,0 +1,92 @@
...
+        }
+        return value == null ? '' : String(value);
+    });
+
+    useEffect(() => {
+        // Seed the grid editor value on mount: ag-grid keeps the last value passed to
+        // onValueChange, and a character that triggered editing (eventKey) does not fire the
+        // input's onChange — without this the typed character would be lost on immediate commit.
+        onValueChange(toNumber(text));
+        refInput.current?.focus();
```
ghazwarhili: .select() on mount makes the editor select the hole value when it opens,

const startedByTyping = eventKey != null && eventKey.length === 1;
    if (!startedByTyping) {
        input?.select();
    }

src/components/ui/csvPicker/csv-picker.tsx
```diff
@@ -0,0 +1,135 @@
...
+                onFileChange(acceptedFile);
+                onFileError(undefined);
+                if (onAppend && onReplace) {
+                    if (hasExistingData?.()) {
+                        setPendingImport({ results, file: acceptedFile });
+                    } else {
+                        onReplace(results, acceptedFile);
+                    }
+                } else {
+                    onComplete?.(results, acceptedFile);
```
ghazwarhili, line 89: onFileChange(acceptedFile);   immediate apply (onComplete mode)
onComplete?.(results, acceptedFile);

src/components/ui/csvPicker/csv-picker.tsx
```diff
@@ -0,0 +1,135 @@
...
+            } else {
+                // Only reflect the file once it is valid: on error the previously selected file name
+                // (and the table data) is kept unchanged.
+                onFileChange(acceptedFile);
+                onFileError(undefined);
+                if (onAppend && onReplace) {
+                    if (hasExistingData?.()) {
+                        setPendingImport({ results, file: acceptedFile });
+                    } else {
+                        onReplace(results, acceptedFile);
```
ghazwarhili, line 85: commit the file name when the import is actually applied 
onFileChange(acceptedFile); 
onReplace(results, acceptedFile);

src/components/ui/csvPicker/csv-picker.tsx
```diff
@@ -0,0 +1,135 @@
...
+                onFileError(
+                    intl.formatMessage(
+                        { id: 'tooManyLinesInCsvFile' },
+                        { value: maxLineNumber, filename: acceptedFile.name }
+                    )
+                );
+            } else {
+                // Only reflect the file once it is valid: on error the previously selected file name
+                // (and the table data) is kept unchanged.
+                onFileChange(acceptedFile);
```
ghazwarhili: removed from here

src/components/ui/csvPicker/csv-picker.tsx
```diff
@@ -0,0 +1,135 @@
...
+                                    width: '100%',
+                                    height: '100%',
+                                }}
+                            />
+                        </Button>
+                    </>
+                )}
+            </CSVReader>
+            {onAppend && onReplace && (
+                <CsvPickerConfirmationDialog
```
ghazwarhili, line 135: the name is committed only upon confirmation by wrapping the dialog handlers:
                    pendingImport={pendingImport}
                    onReplace={(r, f) => {
                        onFileChange(f);
                        onReplace(r, f);
                    }}
                    onAppend={(r, f) => {
                        onFileChange(f);
                        onAppend(r, f);
                    }}
                    onClose={() => setPendingImport(null)}

src/components/ui/csvPicker/csv-picker.tsx
```diff
@@ -0,0 +1,135 @@
...
+    maxLineNumber?: number;
+    disabled?: boolean;
+    language: string;
+    parseConfig?: Partial<ParseConfig<TData>>;
+    selectedFile?: File;
+    onFileChange: (file: File | undefined) => void;
+    onFileError: (error: string | undefined) => void;
+    onComplete?: (results: ParseResult<TData>, file: File) => void;
+    onAppend?: (results: ParseResult<TData>, file: File) => void;
+    onReplace?: (results: ParseResult<TData>, file: File) => void;
```
ghazwarhili: The component has two usage modes:
Simple mode: only onComplete
Append/replace mode: onAppend + onReplace (triggering the confirmation dialog)

cannot pass onAppend without onReplace 
cannot pass both modes at once
cannot omit callbacks entirely

the current typing makes all callbacks optional and independent, the better approach to enhance the type is a union


```suggestion
    type CsvPickerCallbacks<TData> =
    | {
          onComplete: (results: ParseResult<TData>, file: File) => void;
          onAppend?: never;
          onReplace?: never;
      }
    | {
          onComplete?: never;
          onAppend: (results: ParseResult<TData>, file: File) => void;
          onReplace: (results: ParseResult<TData>, file: File) => void;
      };
export type CsvPickerProps<TData = unknown> = {
    label: string;
    header: string[];
    maxLineNumber?: number;
    disabled?: boolean;
    language: string;
    parseConfig?: Partial<ParseConfig<TData>>;
    selectedFile?: File;
    onFileChange: (file: File | undefined) => void;
    onFileError: (error: string | undefined) => void;
    hasExistingData?: () => boolean;
} & CsvPickerCallbacks<TData>;
```

ghazwarhili:
sonar issue to be fixed https://sonarcloud.io/project/issues?severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG&pullRequest=1166&id=gridsuite_commons-ui&open=AZ6TPwRDRHARCSOaf1ZA

src/components/composite/agGridTable/cellEditors/numericEditor.tsx
flomillot: I kept the previous code

```
    private static isCharNumeric(charStr: string | null) {
        // FM : I added ',' and '.'
        return charStr && /\d|,|\./.test(charStr);
    }
```

I did it long time ago and I just added , and . from the code of the documentation of AG Grid. 
Nobody complained about it since a few years, so I guess it's fine. I don't think we can add negative values in the application anyway.

src/components/composite/agGridTable/cellEditors/numericEditor.tsx
flomillot: I don't understand, it's exactly the same.
The second part of the condition is not checked if the first condition is not satisfied. 
If I misunderstood something, can you explain ?

src/components/composite/agGridTable/cellEditors/numericEditor.tsx
flomillot: I don't think it's what we want. Where do we need negative values in the application ?

src/components/composite/agGridTable/CustomAgGridTable.tsx
flomillot: I volontary kept the previous code, but yes okay you're right

flomillot: image

Same for `onRowDataUpdated`
We could fix it by adding TypeScript, but well...

src/components/composite/agGridTable/BottomTableButtons.tsx
flomillot: I will create a function for that.

src/components/ui/csvPicker/csv-picker.tsx
flomillot: The user is not supposed to change the header. Did you encounter a problem with this ?

src/components/ui/csvPicker/csv-picker.tsx
flomillot: Ah yes OK

src/components/ui/csvPicker/csv-picker.tsx
flomillot: You're right, I change it.

src/components/composite/agGridTable/CustomAgGridTable.tsx
ghazwarhili: you can fixed by const [selectedRows, setSelectedRows] = useState<**any[]**>([]);

src/components/composite/agGridTable/cellEditors/numericEditor.tsx
flomillot: But it's already the case with focus, so what's the difference ?

ghazwarhili: https://www.ag-grid.com/react-data-grid/cell-editors/
Concrete case — double click a cell with value 42:
- focus() only → caret sits inside 42, nothing highlighted; typing 7 gives 427/742
- select() → 42 is highlighted; typing 7 replaces it with 7

src/components/composite/agGridTable/cellEditors/numericEditor.tsx
ghazwarhili: My point was mainly about separating the intent
With the early return:
if (event.key.length !== 1) {
  return;
}
we explicitly filter out non-character keys first and then we can focus only on character validation below

src/components/composite/agGridTable/cellEditors/numericEditor.tsx
ghazwarhili: The question isn't really "where in the application", this component lives in commons-ui, which is a shared library consumed by several apps (gridstudy, gridexplore, …). A generic numeric editor shouldn't bake in an assumption tied to one current usage; it should cover the general case, and signed numbers are part of that.

flomillot: It would maybe introduce regression, managing negative values can be tricky, I prefer let it like that for now as its not related to the feature and not necessary for now.

src/components/composite/agGridTable/CustomAgGridTable.tsx
flomillot: OK thanks, I change onRowDataUpdated also then

ghazwarhili (approved):

PR #1174, opened 2026-06-08
refresh , composite modifications

souissimai:


src/features/network-modification-table/network-modifications-table.tsx
```diff
@@ -126,24 +146,78 @@ export function NetworkModificationsTable({
...
+                fetchSubModificationsForExpandedRows(
+                    loadedComposites.map((mod) => mod.uuid),
+                    freshTopLevel,
+                    setComposedModifications,
+                    true
+                );
+            }
+
+            return freshTopLevel;
         });
```
Meklo: I'm afraid this new part makes the useeffect quite hard to read and it's purpose blurry, what do you think of not handling expanded rows state. 

        const prevMods = composedModificationsRef.current;
        // Uuids now at the top level have an authoritative position there. Any stale
        // carried-over child with the same uuid (cut out of a composite, pasted at root)
        // must be stripped, otherwise it renders twice → duplicate row ids / React keys.
        const newTopLevelUuids = new Set(modifications.map((m) => m.uuid));

        // Carry over already-fetched children (avoids an empty flash during the re-fetch),
        // then deep-filter out any uuid that moved to the top level.
        const nextMods = mergeSubModificationsIntoTree(formatToComposedModification(modifications), prevMods).map(
            (mod) =>
                mod.subModifications.length > 0
                    ? { ...mod, subModifications: removeUuidsFromTree(mod.subModifications, newTopLevelUuids) }
                    : mod
        );
        setComposedModifications(nextMods);

        // Re-fetch authoritative children for every composite that already had loaded children,
        // correcting anything stale that was temporarily preserved above.
        // Source of truth: prevMods — nextMods children may have been filtered just above.
        const loadedComposites: ComposedModificationMetadata[] = [];
        findAllLoadedCompositeModifications(prevMods, loadedComposites);
        if (loadedComposites.length > 0) {
            fetchSubModificationsForExpandedRows(
                loadedComposites.map((m) => m.uuid),
                nextMods,
                setComposedModifications,
                true
            );
        }

We simply need to remove rows at root level that were previously nested, this impy to add this function in utils :

    export function removeUuidsFromTree(
    mods: ComposedModificationMetadata[],
    uuidsToRemove: Set<string>
    ): ComposedModificationMetadata[] {
    return mods
    .filter((m) => !uuidsToRemove.has(m.uuid))
    .map((m) =>
    m.subModifications.length > 0
    ? { ...m, subModifications: removeUuidsFromTree(m.subModifications, uuidsToRemove) }
    : m
    );
    }

Do you see any tradoff with this version ?

src/features/network-modification-table/network-modifications-table.tsx
```diff
@@ -94,6 +107,13 @@ export function NetworkModificationsTable({
 
     const [expanded, setExpanded] = useState<ExpandedState>({});
 
+    // Keep a ref so the modifications useEffect can read current expanded state
+    // without being added as a dependency (which would cause infinite loops).
+    const expandedRef = useRef<ExpandedState>(expanded);
+    useEffect(() => {
+        expandedRef.current = expanded;
+    }, [expanded]);
```
Meklo: Could be removed if we don't process expanded state in the useeffect (see comment above)

src/features/network-modification-table/utils.ts
```diff
@@ -240,9 +240,9 @@ export function fetchSubModificationsForExpandedRows(
     getNetworkModificationsFromComposite(uuidsToFetch).then((subModsByUuid) => {
         setMods((prev) =>
             Object.entries(subModsByUuid).reduce((tree, [uuid, subMods]) => {
-                const liveModifications = formatToComposedModification(subMods.filter((m) => !m.stashed));
                 // Preserve already-loaded children of any nested composites within the new sub-list
                 const existingMod = findModificationInTree(uuid, tree);
+                const liveModifications = formatToComposedModification(subMods.filter((m) => !m.stashed));
```
Meklo: Why this move ? Could be reverted no ?

src/features/network-modification-table/network-modifications-table.tsx
```diff
@@ -65,6 +64,20 @@ interface NetworkModificationsTableProps extends Omit<NetworkModificationEditorN
...
+function collectAllUuidsInTree(mods: ComposedModificationMetadata[]): Set<string> {
+    const uuids = new Set<string>();
+    const visit = (nodes: ComposedModificationMetadata[]) =>
+        nodes.forEach((m) => {
+            uuids.add(m.uuid);
+            visit(m.subModifications);
+        });
+    visit(mods);
+    return uuids;
+}
```
Meklo: To be moved in utils ?

Meklo (changes requested):
Minor remarks and a rewrite suggestion for the useffect

Meklo (approved):

PR #1177, opened 2026-06-09
feat(ByFilterAssignment): add reactive power measurement attributes for static var compensator

dbraquart:
## PR Summary
Because there is currently no static var compensator modification in GridSuite, we add the possibility to change the reactive power measurement value/validity with the "Modify by filter" exisiting modification.


benrejebmoh:
code ok
test ok

benrejebmoh (approved):

PR #1182, opened 2026-06-10
add operator EXISTS/NOT_EXISTS for property expert rule

EtienneLt:


src/components/composite/filter/expert/expertFilterUtils.ts
```diff
@@ -445,18 +445,23 @@ export const queryValidator: QueryValidator = (query) => {
...
-                valid: false,
-                reasons: [RULES.EMPTY_RULE],
-            };
+        } else if (rule.id && dataType === DataType.PROPERTY) {
+            if (isBlankOrEmpty(rule.value?.propertyName) || isBlankOrEmpty(rule.value?.propertyOperator)) {
+                result[rule.id] = {
+                    valid: false,
+                    reasons: [RULES.EMPTY_RULE],
+                };
+            } else if (rule.value?.propertyOperator === 'EXISTS' || rule.value?.propertyOperator === 'NOT_EXISTS') {
```
flomillot:
```suggestion
            } else if (rule.value?.propertyOperator === OPERATOR_OPTIONS.EXISTS.name || rule.value?.propertyOperator === OPERATOR_OPTIONS.NOT_EXISTS.name) {
```

src/components/composite/filter/expert/expertFilterUtils.ts
```diff
@@ -445,18 +445,23 @@ export const queryValidator: QueryValidator = (query) => {
...
+                result[rule.id] = {
+                    valid: true,
+                    reasons: undefined,
+                };
```
flomillot: It's not related to this pull request but I don't understand why we need to put everything in `value` and not use the react-querybuilder native fields, in particular `operator`.

src/components/composite/reactQueryBuilder/PropertyValueEditor.tsx
```diff
@@ -27,6 +32,7 @@ export function PropertyValueEditor(props: ExpertFilterPropertyProps) {
     const { equipmentType, valueEditorProps } = props;
     const valid = useValid(valueEditorProps);
     const intl = useIntl();
+    const customFilterOptions = useCustomFilterOptions();
```
flomillot: Tell me if I'm wrong, but this hook seems not only useless but counter-productive, maybe transform it into a basic function ?

src/components/composite/filter/expert/expertFilterUtils.ts
flomillot: Or `OPERATOR_OPTIONS.EXISTS.customName` If I understood well.

flomillot (approved):

PR #1186, opened 2026-06-11
Migrate security analysis n-k results component and its dependencies to commons-ui

FranckLecuyer:


src/features/results/securityanalysis/index.ts
```diff
@@ -0,0 +1,34 @@
...
+    ContingenciesFromConstraintItem,
+    CutOffPowerFromConstraintsItem,
+    LimitViolation,
+    Constraint,
+    ContingencyItem,
+    Contingency,
+    ConnectivityResult,
+    PreContingencyResult,
+} from './security-analysis.type';
+export { NmkType } from './security-analysis.type';
```
klesaulnier: replace all by "export * from ..." ?

src/features/results/securityanalysis/security-analysis-result-nmk.tsx
```diff
@@ -0,0 +1,114 @@
...
+        [getRowStyle, nmkType]
+    );
+
+    const overlayNoRowsTemplate = getNoRowsMessage(
+        resultStatusMessages,
+        rows,
+        securityAnalysisStatus,
+        !isLoadingResult
+    );
+
```
klesaulnier, line 98: if easy to do, move this to src/components/results/securityanalysis/security-analysis-result-tab.tsx is gridstudy-app
it would factorize the code between n and nmk result
+ this component only needs "overlayNoRowsTemplate" as prop, it does not have to be linked to resultStatusMessages, securityAnalysisStatus and isLoadingResult

Same thing for src/components/results/securityanalysis/security-analysis-result-n.tsx in gridstudy-app

src/features/results/securityanalysis/security-analysis.type.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export type SecurityAnalysisNmkResult = {
+    content?: (
+        | ContingenciesFromConstraintItem[]
+        | ConstraintsFromContingencyItem[]
+        | CutOffPowerFromConstraintsItem[]
+        | null
+    )[];
+    [key: string]: unknown;
+};
```
klesaulnier, line 110: we had a "Page" wrapper for this one, we could use it here ?
Maybe it's too complicated to migrate Page here ?

src/features/results/securityanalysis/security-analysis.type.ts
```diff
@@ -0,0 +1,153 @@
...
+export interface Contingency {
+    contingency?: ContingencyItem;
+    limitViolation?: LimitViolation;
+}
+
+export interface PreContingencyResult {
+    subjectId?: string;
+    status: string;
+    limitViolation?: LimitViolation;
+}
```
klesaulnier: as stated in gridstudy PR, I think it should stay in gridstudy-app, since we won't migrate N results here

klesaulnier (commented):
Tests: OK
Some change requests

src/features/results/securityanalysis/index.ts
FranckLecuyer: :+1: 
Done

src/features/results/securityanalysis/security-analysis.type.ts
FranckLecuyer: :+1: 
I put it back in gridstudy

src/features/results/securityanalysis/security-analysis-result-nmk.tsx
FranckLecuyer: To much work here : 

rows is computed inside this component SecurityAnalysisResultNmk ...
and securityAnalysisStatus is also used in this component to compute rowsToShow ...

So, i think it is better to do this in another PR ...

src/features/results/securityanalysis/security-analysis.type.ts
FranckLecuyer: In another PR, i think ...

klesaulnier (approved):

PR #1197, opened 2026-06-18
Parameters new design

khouadrired:
This PR standardizes the layout and action handling across all 10 parameter categories in the application. It migrates all parameter forms to a unified `ParameterLayout` component and replaces the legacy `renderActions` pattern with a centralized `ParameterActions` interface.

#### Key Changes
- **Unified Layout**: Refactored all parameter forms (Load Flow, Security Analysis, Network Visualizations, etc.) to use `ParameterLayout` for consistent scrolling behavior and responsive design.
- **Centralized Actions**: Introduced a `ParameterActions` object to group standard callbacks (Save, Reset, Validate, Choose Settings) and an `extra` prop for custom dialogs/modals.
- **Responsive Improvements**: Fixed XS breakpoint layout issues in `ParametersTabs` to ensure content remains visible when using the top-level selection component.
- **Code Cleanup**: Removed legacy CSS hacks (e.g., hardcoded `maxHeight` and absolute positioning) 

#### Benefits
All parameter forms now share the same structure: Header -> Scrollable Content -> Sticky Footer.
Reduced boilerplate by passing a single `actions` object instead of multiple individual props.


src/features/parameters/common/parameter-layout.tsx
```diff
@@ -0,0 +1,128 @@
...
+                            {preFillOnClick && (
+                                <LabelledButton
+                                    callback={preFillOnClick}
+                                    label="button.prefill"
+                                    data-testid="PrefillButton"
+                                    startIcon={<Upload />}
+                                />
+                            )}
+                            {resetOnClick && (
+                                <LabelledButton
```
FranckLecuyer: Missing required tooltip

FranckLecuyer (changes requested):

src/features/parameters/common/parameter-layout.tsx
khouadrired: added

src/translations/en/parameters.ts
```diff
@@ -473,4 +473,7 @@ export const parametersEn = {
     NODE_BUILT: 'Node built',
     NODE_NOT_BUILT: 'Node not built',
     ROOT_NODE: 'Root node',
+
+    //Tooltips
+    'tooltip.reset': 'Paramètre par défaut du profil utilisateur',
```
FranckLecuyer: To be translated

FranckLecuyer (changes requested):

src/translations/en/parameters.ts
khouadrired: done

FranckLecuyer (approved):

src/features/parameters/common/parameter-layout.tsx
```diff
@@ -0,0 +1,130 @@
...
+                    </Grid>
+                </Grid>
+            </Box>
+            {header && <Box sx={styles.header}>{header}</Box>}
+            <Box sx={[styles.content, contentSx]}>{isLoading ? <LinearProgress /> : children}</Box>
+            <Box sx={styles.footer}>
+                {saveOnClick && <LabelledButton label="save" data-testid="SaveButton" callback={saveOnClick} />}
+                {validateOnClick && (
+                    <SubmitButton variant="contained" data-testid="ValidateButton" onClick={validateOnClick}>
+                        <FormattedMessage id="validate" />
```
TheMaskedTurtle: SubmitButton already contains this

src/features/parameters/common/parameter-layout/parameter-layout.tsx
```diff
@@ -0,0 +1,130 @@
...
+    isLoading,
+    contentSx,
+    actions,
+}: Readonly<ParameterLayoutProps>) {
+    const { preFillOnClick, resetOnClick, saveOnClick, validateOnClick } = actions ?? {};
+
+    const { isXsScreen } = useParameterLayoutContext();
+
+    return (
+        <Stack sx={styles.stack}>
```
TheMaskedTurtle, line 91: We usually don't use Stack, for homogeneity I would rather use <Grid direction"column">

src/features/parameters/common/parameter-layout.tsx
```diff
@@ -0,0 +1,130 @@
+/**
+ * Copyright (c) 2026, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { Box, Grid, LinearProgress, Stack, ButtonGroup, Tooltip } from '@mui/material';
```
TheMaskedTurtle: Why not migrating to Grid2 directly ?

src/features/parameters/common/parameter-layout.tsx
```diff
@@ -0,0 +1,130 @@
...
+    validateOnClick?: React.MouseEventHandler<HTMLButtonElement>;
+    extra?: ReactNode;
+}
+
+interface ParameterLayoutProps {
+    children: ReactNode;
+    title?: string;
+    header?: ReactNode;
+    isLoading?: boolean;
+    contentSx?: any;
```
TheMaskedTurtle: This is weird according to me, why not handled by the content itself ?

src/features/parameters/dynamic-margin-calculation/dynamic-margin-calculation-form.tsx
```diff
@@ -22,64 +22,78 @@ import { UseComputationParametersFormReturn } from '../common/utils';
...
-        </>
+        <ParameterLayout
+            title="DynamicMarginCalculation"
+            header={
+                <>
+                    {renderTitleFields?.()}
+                    {paramsLoaded && (
+                        <>
+                            <ProviderParam options={formattedProviders} />
+                            <Tabs
```
TheMaskedTurtle: Should we really put Tabs in the header ?

TheMaskedTurtle (changes requested):
- I can still see a lot of custom sx here and there. It would be a great occasion to rationalize that and to limit them to what is really useful. It leads to a lot of little differences between screens, so the less we have them the better we stand. So, try to limit them as much as possible.
- I think you should use Grid2 directly with import { Grid2 as Grid } to avoid second time refactoring, and rather use Grid than Stack or Box, for project consistency. 
- I am not a big fan of putting a lot of things in the header section, for me it should be limited to the title. If we want something specific for the provider I would rather add a specific section provider in the parameter layout, instead of using the header one. Otherwise I would just let it be handled by the content itself.
- I am also surprised by the renderTitleFields present in the forms, not sure it should be here, but maybe that would be something to dig deeper int the future.

Could you add more description in the PR about which functionality it covers and add some screenshots of the changes.

Thanks 🙏

src/features/parameters/common/parameter-layout/parameter-layout.tsx
khouadrired: Regarding MUI v6, if we switch to Grid2, the "direction="column"" attribute will no longer be supported; therefore, it is recommended to use "Stack". 
https://v6.mui.com/material-ui/react-grid2/#column-direction

src/features/parameters/common/parameter-layout.tsx
khouadrired: i tried to do that , but the content of parameters got messed up , as we will have legacy Grid inside a GridV2 container

src/features/parameters/dynamic-margin-calculation/dynamic-margin-calculation-form.tsx
khouadrired: it was just a design choice to have a fixed layout , but we really don't need to.

src/features/parameters/common/parameter-layout.tsx
khouadrired: you mean the contentSx ?

src/features/parameters/common/parameter-layout.tsx
khouadrired: will check that.

src/features/parameters/common/parameter-layout.tsx
TheMaskedTurtle: ok, too bad :/

src/features/parameters/common/parameter-layout.tsx
khouadrired: fixed

src/features/parameters/common/parameter-layout.tsx
khouadrired: fixed

src/features/parameters/dynamic-margin-calculation/dynamic-margin-calculation-form.tsx
khouadrired: my idea to put tabs in the header was originaly from a layout prespective , as the tabs can also be a fixed component.

TheMaskedTurtle (changes requested):
According to me, there is a misconception about components hierarchy here. The *Form components should contain the whole form, provider and tabs included, and it should be agnostic of the surrounding environment. Then, in gridstudy-app the Tab content should be your newly introduced Layout with the Form as content. In gridexplore-app, it should be a MuiDialog with inside stacked NameElementEditorForm, followed by the parameter Form. Here it is the form  that contains the layout, and the inline that contains the form, it causes many optional fields in your layout to match both behaviors when in reality it is not used in gridexplore-app. We can talk further about this, but I think it is really important to reverse the logic.

TheMaskedTurtle:
Is there a plan to use the ParameterLayoutContext in the future ? For now, I feel like it is not used and it might be a little over-engineered, maybe we can introduce it only when it will be used.

thangqp (commented):
**XS screen**
Image

In xs size, IMO, the dropbox should be the same line of toolbars Prefill/Reset to reduce the usage in vertical.

**Normal screen**
image

We should consider the dropbox as a replacement of the label that you show in normal screen. To see with Stephane, I dont found these buttons in his Figma files.

thangqp:
> According to me, there is a misconception about components hierarchy here. The *Form components should contain the whole form, provider and tabs included, and it should be agnostic of the surrounding environment. Then, in gridstudy-app the Tab content should be your newly introduced Layout with the Form as content. In gridexplore-app, it should be a MuiDialog with inside stacked NameElementEditorForm, followed by the parameter Form. Here it is the form that contains the layout, and the inline that contains the form, it causes many optional fields in your layout to match both behaviors when in reality it is not used in gridexplore-app. We can talk further about this, but I think it is really important to reverse the logic.

Agreed! Form should only contain functional fields and not to be shipped with a layout with alot of optional inputs..
I see it is possible with dynawo params in which CustomFormProvider lives in xxx-inline components

thangqp:
> Is there a plan to use the ParameterLayoutContext in the future ? For now, I feel like it is not used and it might be a little over-engineered, maybe we can introduce it only when it will be used.

I found it is used only at one place to show/hide computation type here : https://github.com/gridsuite/commons-ui/pull/1197/changes#diff-9d2eedc91893a01e3c7b6b2d377f8018e02c586ede4340b7de3d5671a0c294ddR85
But I am not sure it is a demand in the marquette of Stephane

src/features/parameters/common/parameter-layout.tsx
```diff
@@ -0,0 +1,128 @@
...
+export function ParameterLayoutProvider({
+    isXsScreen,
+    children,
+}: Readonly<{
+    isXsScreen: boolean;
+    children: ReactNode;
+}>) {
+    const isXsScreenMemo = useMemo(() => ({ isXsScreen }), [isXsScreen]);
+    return <ParameterLayoutContext.Provider value={isXsScreenMemo}>{children}</ParameterLayoutContext.Provider>;
+}
```
TheMaskedTurtle: The whole application is not adapted to small formats. For me that's over-engineered to think about that only for this specifc case when everything else will not be adapted anyway. It's ok to keep it, but IMO it just brings noise for now.

src/features/parameters/common/parameter-layout.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export interface ParameterActions {
+    preFillOnClick?: React.MouseEventHandler<HTMLButtonElement>;
+    resetOnClick?: React.MouseEventHandler<HTMLButtonElement>;
+    saveOnClick?: React.MouseEventHandler<HTMLButtonElement>;
+    validateOnClick?: React.MouseEventHandler<HTMLButtonElement>;
+    validateDisabled?: boolean;
+    extra?: ReactNode;
```
TheMaskedTurtle: I think this is not necessary, you're using it only for saving or confirmation dialogs for now. So for me, it can just be declared in the Inline component directly, all the logic is already handled in this component anyway. In the layout, as you did, you just need the callbacks to handle clicks

src/features/parameters/common/parameter-layout.tsx
```diff
@@ -0,0 +1,128 @@
...
+    content: {
+        flexGrow: 1,
+        overflowY: 'auto',
+        overflowX: 'hidden',
+        minHeight: 0, // Critical for flex-grow with overflow
+    },
+    footer: {
+        flexShrink: 0,
+        p: 1,
+        borderTop: (theme: Theme) => `1px solid ${theme.palette.divider}`,
```
TheMaskedTurtle: We have a LineSeparator component

src/features/parameters/loadflow/load-flow-parameters-form.tsx
```diff
@@ -67,33 +56,26 @@ export function LoadFlowParametersForm({
...
-                            />
-                        </Grid>
-                    </Box>
-                ) : (
-                    <LinearProgress />
-                )}
-                {renderActions && <Box sx={styles.actions}>{renderActions()}</Box>}
-            </Box>
-        </CustomFormProvider>
+        <Grid container sx={styles.content}>
```
TheMaskedTurtle: It feeld like a `direction="column"` is missing here

src/features/parameters/loadflow/load-flow-parameters-form.tsx
```diff
@@ -67,33 +56,26 @@ export function LoadFlowParametersForm({
...
+            {renderTitleFields?.()}
```
TheMaskedTurtle: Now this is only used in Dialog it would be nice to remove it from props and just use it in dialog
```
<CustomMuiDialog
            open={open}
            onClose={onClose}
            onSave={loadflowMethods.onSaveDialog}
            onValidationError={loadflowMethods.onValidationError}
            formContext={{
                ...loadflowMethods.formMethods,
                validationSchema: loadflowMethods.formSchema,
                removeOptional: true,
                language,
            }}
            titleId={titleId}
            disabledSave={disableSave}
        >
            <LoadFlowProvider>
                <NameElementEditorForm
                    initialElementName={name}
                    activeDirectory={activeDirectory}
                    elementType={ElementType.LOADFLOW_PARAMETERS}
                />
                <LoadFlowParametersForm loadflowMethods={loadflowMethods} />
            </LoadFlowProvider>
        </CustomMuiDialog>
```

TheMaskedTurtle (changes requested):
- It needs to be tested extensively, to be sure all layouts in both gridexplore and gridstudy are clean, I haven't done it but just on loadflow parameters I noticed regressions. 
- Still not convinced by the "small screen handling".
- Except for the small points I raised (about `extra` prop and NameElementEditorForm), I think in terms of component hierarchy and responsabilities, it is perfect like that, thanks for the rework :)

TheMaskedTurtle:
Capture d’écran du 2026-06-25 16-00-38

Screenshot to prove my point about small screens, it's not like it's really useable anyway ^^

src/features/parameters/common/parameter-layout.tsx
khouadrired: will try to update it

src/features/parameters/common/parameter-layout.tsx
khouadrired: we will make use of this provider in the next steps of the design, it will be used for the Grid system .

src/components/ui/reactHookForm/text/UniqueNameInput.tsx
```diff
@@ -81,6 +82,9 @@ export function UniqueNameInput({
         </InputAdornment>
     );
 
+    console.log('value : ', value);
+    console.log('initial : ', currentName);
```
TheMaskedTurtle: Should remove

src/features/parameters/dynamic-margin-calculation/dynamic-margin-calculation-form.tsx
```diff
@@ -22,64 +19,45 @@ import { UseComputationParametersFormReturn } from '../common/utils';
...
 
-    const { selectedTab, tabsWithError, onTabChange, onError } = useTabs({
+    const { selectedTab, tabsWithError, onTabChange } = useTabs({
         defaultTab: Object.values(TabValues)[0],
         tabEnum: TabValues,
     });
+
+    if (!paramsLoaded) {
+        return <LinearProgress />;
+    }
```
TheMaskedTurtle: Isn't it handled in the layout ?

src/features/parameters/common/parameter-layout.tsx
```diff
@@ -0,0 +1,209 @@
...
+        setOpenResetConfirmation(false);
+    }, []);
+
+    const intl = useIntl();
+
+    return (
+        <Stack sx={styles.stack}>
+            {(title || selectParameterHandler || resetHandler) && (
+                <Box sx={styles.title}>
+                    <Grid container justifyContent="space-between" sx={{ alignItems: 'center' }}>
```
TheMaskedTurtle: You're mixing sx props and inlined props, I think it should be harmonized

src/features/parameters/loadflow/load-flow-parameters-dialog.tsx
```diff
@@ -70,19 +71,14 @@ export function LoadFlowParametersEditionDialog({
...
-                            <NameElementEditorForm
-                                initialElementName={name}
-                                activeDirectory={activeDirectory}
-                                elementType={ElementType.LOADFLOW_PARAMETERS}
-                            />
-                        );
-                    }}
-                />
+                <Grid container sx={{ width: '100%' }}>
+                    <NameElementEditorForm
```
TheMaskedTurtle, line 75: You migrated this only for loadflow but it has to be done almost for all parameters

src/features/parameters/voltage-init/voltage-init-parameters-inline.tsx
```diff
@@ -79,86 +69,34 @@ export function VoltageInitParametersInLine({
...
-        />
+        <CustomFormProvider
+            validationSchema={voltageInitMethods.formSchema}
+            {...voltageInitMethods.formMethods}
+            removeOptional
+        >
+            <ParameterLayout
+                title="VoltageInit"
+                isLoading={voltageInitMethods.paramsLoading}
+                parameterType={ElementType.SENSITIVITY_PARAMETERS}
```
TheMaskedTurtle: ElementType.VOLTAGE_INIT_PARAMETERS

src/components/ui/reactHookForm/text/UniqueNameInput.tsx
```diff
@@ -53,6 +53,7 @@ export function UniqueNameInput({
         fieldState: { error },
     } = useController({
         name,
+        defaultValue: currentName,
```
TheMaskedTurtle: Is that useful ?

src/features/parameters/loadflow/load-flow-parameters-form.tsx
```diff
@@ -67,33 +54,29 @@ export function LoadFlowParametersForm({
...
-                            />
-                        </Grid>
-                    </Box>
-                ) : (
-                    <LinearProgress />
-                )}
-                {renderActions && <Box sx={styles.actions}>{renderActions()}</Box>}
-            </Box>
-        </CustomFormProvider>
+        <Grid container sx={styles.content}>
```
TheMaskedTurtle: We should not have extra sx here I think.

TheMaskedTurtle: Don't forget to clean, you're using only content from styles now (and I think it can be removed too) but they are still declared

src/features/parameters/common/parameter-layout/parameter-layout.tsx
```diff
@@ -0,0 +1,209 @@
...
+    isXsScreen: boolean;
+    children: ReactNode;
+}>) {
+    const isXsScreenMemo = useMemo(() => ({ isXsScreen }), [isXsScreen]);
+    return <ParameterLayoutContext.Provider value={isXsScreenMemo}>{children}</ParameterLayoutContext.Provider>;
+}
+
+export type CreateParameterDialogConfig<T extends FieldValues> = {
+    studyUuid: UUID | null;
+    getParameterValues: UseFormGetValues<T> | any;
```
TheMaskedTurtle: you should fix that warning

src/features/parameters/loadflow/load-flow-parameters-inline.tsx
```diff
@@ -90,87 +64,28 @@ export function LoadFlowParametersInline({
...
-            />
+            >
+                <ParameterLayout
+                    title="LoadFlow"
+                    isLoading={!loadflowMethods.paramsLoaded}
+                    parameterType={ElementType.LOADFLOW_PARAMETERS}
+                    createParameter={{
+                        studyUuid,
+                        getParameterValues: () => loadflowMethods.formatNewParams(getValues()),
+                        parameterFormatter: (newParams) => newParams,
```
TheMaskedTurtle: Either you really split getter and formatter then it should more look like :
```
getParameterValues: getValues,
parameterFormatter: (values) => loadflowMethods.formatNewParams(values),
```
or you just use getParameterValues with the full method as you did but then no need to define parameterFormatter

src/features/parameters/dynamic-margin-calculation/dynamic-margin-calculation-form.tsx
khouadrired: no , it will work also for inline form

src/components/ui/reactHookForm/text/UniqueNameInput.tsx
khouadrired: for inline the name is rendered before the form so having that as a value fixes it

src/features/parameters/common/parameter-layout.tsx
khouadrired: will do that

src/components/ui/reactHookForm/text/UniqueNameInput.tsx
khouadrired: will do that

khouadrired: will do that

src/features/parameters/loadflow/load-flow-parameters-dialog.tsx
khouadrired: i'm checking that

src/features/parameters/common/parameter-layout/parameter-layout.tsx
khouadrired: i used the same param type in that we pass to dialogs

src/features/parameters/loadflow/load-flow-parameters-inline.tsx
khouadrired: i will check

src/components/ui/reactHookForm/text/UniqueNameInput.tsx
TheMaskedTurtle: I don't get it, this is not used in inline. It's just for dialogs.

src/features/parameters/common/parameter-layout/parameter-layout.tsx
```diff
@@ -0,0 +1,190 @@
...
+                                            startIcon={<RestartAlt />}
+                                        />
+                                    </Tooltip>
+                                )}
+                            </ButtonGroup>
+                        </Grid>
+                    </Grid>
+                </Box>
+            )}
+            <Box sx={styles.content}>{isLoading ? <LinearProgress /> : children}</Box>
```
TheMaskedTurtle, line 121: If you're handling loading behavior in the forms you should remove it from here.

src/features/parameters/common/parameter-layout/parameter-layout.tsx
```diff
@@ -0,0 +1,190 @@
...
+                </Grid>
+                {createParameter && <LabelledButton label="save" data-testid="SaveButton" callback={handleSaveClick} />}
+                {validateHandler && (
+                    <SubmitButton
+                        variant="contained"
+                        data-testid="ValidateButton"
+                        onClick={validateHandler}
+                        disabled={validateDisabled}
+                    />
+                )}
```
TheMaskedTurtle:
```suggestion
            <Stack spacing={1}>
                    <LineSeparator />
                    <Grid container spacing={1}>
                        {createParameter && (
                            <Grid item>
                                <LabelledButton label="save" data-testid="SaveButton" callback={handleSaveClick} />
                            </Grid>
                        )}
                        {validateHandler && (
                            <Grid item>
                                <SubmitButton
                                    variant="contained"
                                    data-testid="ValidateButton"
                                    onClick={validateHandler}
                                    disabled={validateDisabled}
                                />
                            </Grid>
                        )}
                    </Grid>
                </Stack>
```
This part of the layout feels awkward. For me, something like this instead.

src/features/parameters/common/parameter-layout/parameter-layout.tsx
```diff
@@ -0,0 +1,190 @@
...
+                        message="resetParamsConfirmation"
+                        validateButtonLabel="validate"
+                        openConfirmationPopup={openResetConfirmation}
+                        setOpenConfirmationPopup={handleCancelReset}
+                        handlePopupConfirmation={() => {
+                            resetHandler();
+                            setOpenResetConfirmation(false);
+                        }}
+                    />
+                )}
```
TheMaskedTurtle: You can put this out of the stack to make it clearer like 
```
<>
    <Stack>
    </Stack>
    {all the dialogs}
</>
```

src/utils/types/elementType.ts
```diff
@@ -25,6 +25,7 @@ export enum ElementType {
     DYNAMIC_SECURITY_ANALYSIS_PARAMETERS = 'DYNAMIC_SECURITY_ANALYSIS_PARAMETERS',
     DYNAMIC_MARGIN_CALCULATION_PARAMETERS = 'DYNAMIC_MARGIN_CALCULATION_PARAMETERS',
     NETWORK_VISUALIZATIONS_PARAMETERS = 'NETWORK_VISUALIZATIONS_PARAMETERS',
+    STATE_ETIMATION_PARAMETERS = 'STATE_ETIMATION_PARAMETERS',
```
TheMaskedTurtle: Typo

src/translations/fr/parameters.ts
```diff
@@ -496,4 +496,7 @@ export const parametersFr = {
     NODE_BUILT: 'Nœud réalisé',
     NODE_NOT_BUILT: 'Nœud non réalisé',
     ROOT_NODE: 'Nœud racine',
+
+    // Tooltips
+    'tooltip.reset': 'Paramètre par défaut du profil utilisateur',
```
TheMaskedTurtle: typo: paramètres

src/translations/en/parameters.ts
```diff
@@ -473,4 +473,7 @@ export const parametersEn = {
     NODE_BUILT: 'Node built',
     NODE_NOT_BUILT: 'Node not built',
     ROOT_NODE: 'Root node',
+
+    // Tooltips
+    'tooltip.reset': 'User default parameters',
```
TheMaskedTurtle, line 478: Default parameters from user profile ?

src/components/ui/reactHookForm/text/UniqueNameInput.tsx
khouadrired: > I don't get it, this is not used in inline. It's just for dialogs.

with the new setup in the dialogs , the initial value of the nameinput is null. it should have the current config name.

thangqp: Is there a bug here? I have tested without new code but i dont see any problem.

src/features/parameters/short-circuit/short-circuit-parameters-inline.tsx
```diff
@@ -66,93 +55,35 @@ export function ShortCircuitParametersInLine({
...
+            removeOptional
+        >
+            <ParameterLayout
+                title="ShortCircuit"
+                isLoading={!shortCircuitMethods.paramsLoaded}
+                parameterType={ElementType.SHORT_CIRCUIT_PARAMETERS}
+                createParameter={{
+                    studyUuid,
+                    getParameterValues: () => shortCircuitMethods.formatNewParams(getValues()),
+                    parameterFormatter: (newParams) => newParams,
```
thangqp: rhf getter vs  dto formatter

thangqp (approved):
Good for me

src/features/parameters/loadflow/load-flow-parameters-inline.tsx
thangqp: https://github.com/gridsuite/commons-ui/pull/1197#discussion_r3514877540

src/features/parameters/dynamic-margin-calculation/dynamic-margin-calculation-form.tsx
thangqp: IMO, it should be controlled inside each dialog, similary to what we haved done in the inline by passing the loading to the layout through layout props.

src/features/parameters/common/parameter-layout/parameter-layout.tsx
```diff
@@ -0,0 +1,190 @@
...
+    validateHandler?: React.MouseEventHandler<HTMLButtonElement>;
+    validateDisabled?: boolean;
+}
+
+const styles = {
+    stack: {
+        height: '100%',
+        overflow: 'hidden',
+        position: 'relative',
+        display: 'flex',
```
TheMaskedTurtle: I think these are useless

src/features/parameters/common/parameter-layout/parameter-layout.tsx
```diff
@@ -0,0 +1,190 @@
...
+    },
+    title: {
+        flexShrink: 0,
```
TheMaskedTurtle: Not sure it's usefull

src/features/parameters/common/parameter-layout/parameter-layout.tsx
```diff
@@ -0,0 +1,190 @@
...
+        paddingBottom: 2,
+    },
+    content: {
+        flexGrow: 1,
+        overflowY: 'auto',
+        overflowX: 'hidden',
+        minHeight: 0, // Critical for flex-grow with overflow
```
TheMaskedTurtle: Same, does not fell useful

src/features/parameters/common/parameter-layout/parameter-layout.tsx
```diff
@@ -0,0 +1,190 @@
...
+    },
+    footer: {
+        flexShrink: 0,
```
TheMaskedTurtle: Same

src/features/parameters/dynamic-margin-calculation/dynamic-margin-calculation-form.tsx
TheMaskedTurtle: I also find it a bit counter-intuitive to handle that in the form itself

src/features/parameters/loadflow/load-flow-parameters-inline.tsx
khouadrired: Done

src/features/parameters/loadflow/load-flow-parameters-form.tsx
khouadrired: Done

src/features/parameters/voltage-init/voltage-init-parameters-inline.tsx
khouadrired: Fixed

src/features/parameters/loadflow/load-flow-parameters-dialog.tsx
khouadrired: Done

src/features/parameters/common/parameter-layout.tsx
khouadrired: Done

src/components/ui/reactHookForm/text/UniqueNameInput.tsx
TheMaskedTurtle: I tested without the fix and I don't see any issue

src/features/parameters/short-circuit/short-circuit-parameters-inline.tsx
khouadrired: fixed

src/features/parameters/common/parameter-layout/parameter-layout.tsx
khouadrired: done , removed from the forms

src/features/parameters/dynamic-margin-calculation/dynamic-margin-calculation-form.tsx
khouadrired: fixed

src/components/ui/reactHookForm/text/UniqueNameInput.tsx
khouadrired: restored , don't get the issue anymore

src/translations/en/parameters.ts
khouadrired: we have a requirement to put `User defaut parameters`.

src/features/parameters/common/parameter-layout/parameter-layout.tsx
khouadrired: done

src/features/parameters/common/parameter-layout/parameter-layout.tsx
TheMaskedTurtle: The last three survivors ? @khouadrired 😄

TheMaskedTurtle (approved):
Ok for the architecture. Thanks for the refactoring ! 

I still think someone from your team should test all of this very carefully again, especially on the error management in tabs, I think there is a risk. I haven't performed comprehensive functionnal tests, I just approve the core concept of the refactoring.

thangqp:
> Ok for the architecture. Thanks for the refactoring !
> 
> I still think someone from your team should test all of this very carefully again, especially on the error management in tabs, I think there is a risk. I haven't performed comprehensive functionnal tests, I just approve the core concept of the refactoring.

This PR is not merge before the migration Prs Grid -> Grid2. And then, they layout component will be migrated to Grid2 before retesting then merge..

src/features/parameters/common/parameter-layout/parameter-layout.tsx
khouadrired: done

PR #1201, opened 2026-06-19
ihm shared modif

souissimai:


src/features/network-modification-table/renderers/cell-renderers.tsx
```diff
@@ -74,6 +76,14 @@ export function DescriptionCellRenderer({ row, table }: CCtx) {
     );
 }
 
+export function ShareCellRenderer({ row, table }: CCtx) {
```
Mathieu-Deharbe: Using "Share" (ie the verb), gives the impression that the code allows to share something. While it only displays a reference to a shared modification.

In the back this is a ModificationReference so I would maybe call this `ReferenceCellRenderer`.

src/features/network-modification-table/renderers/reference-link-cell.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { useSnackMessage } from '../../../hooks';
+import { fetchAppsMetadata, fetchNetworkModification } from '../../../services';
+import { ComposedModificationMetadata, ExploreMetadata, Metadata, snackWithFallback } from '../../../utils';
+import { CustomMenuItem } from '../../../components';
+import { ReferenceModificationInfos } from '../utils';
+import { DatasetLinkedIcon } from '../../../components/ui/icons/DatasetLinkedIcon';
+
+export interface ReferenceLinkCellProps {
+    data: ComposedModificationMetadata;
+    isDisabled?: boolean;
```
Mathieu-Deharbe: A technical discussion in teams concluded that we should use `disabled` instead of `isDisabled` in order to imitate MUI.

```suggestion
    disabled?: boolean;
```

src/features/network-modification-table/renderers/cell-renderers.tsx
```diff
@@ -74,6 +76,14 @@ export function DescriptionCellRenderer({ row, table }: CCtx) {
...
+    const { meta } = table.options;
+
+    // A shared-modification (reference) row shows a "copy link" action instead of the description note.
```
Mathieu-Deharbe: I am not sure that this comment is needed. But the end definitely is not.
```suggestion
    // A reference to a shared-modification row shows a "copy link" action
```

src/features/network-modification-table/renderers/reference-link-cell.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { LinkRounded as LinkRoundedIcon } from '@mui/icons-material';
+import { useSnackMessage } from '../../../hooks';
+import { fetchAppsMetadata, fetchNetworkModification } from '../../../services';
+import { ComposedModificationMetadata, ExploreMetadata, Metadata, snackWithFallback } from '../../../utils';
+import { CustomMenuItem } from '../../../components';
+import { ReferenceModificationInfos } from '../utils';
+import { DatasetLinkedIcon } from '../../../components/ui/icons/DatasetLinkedIcon';
+
+export interface ReferenceLinkCellProps {
+    data: ComposedModificationMetadata;
```
Mathieu-Deharbe, line 18: It looks like only the uuid is useful. Maybe you should only store the uuid and pass only it as a prop instead of a whole ComposedModificationMetadata ?

src/features/network-modification-table/renderers/reference-link-cell.tsx
```diff
@@ -0,0 +1,104 @@
...
+    isDisabled?: boolean;
+}
+
+export function ReferenceLinkCell({ data, isDisabled = false }: Readonly<ReferenceLinkCellProps>) {
```
Mathieu-Deharbe: I don't know if this is important but in the specs the menu is at the bottom right of the cursor while here it is under it : 

Image

src/features/network-modification-table/renderers/reference-link-cell.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const [isLoading, setIsLoading] = useState(false);
+    const open = Boolean(anchorEl);
+
+    const handleOpen = useCallback((event: React.MouseEvent<HTMLElement>) => {
+        // prevent the row click handler (edit dialog) from firing
+        event.stopPropagation();
+        setAnchorEl(event.currentTarget);
+    }, []);
+
+    // For MUI Menu's onClose (event is typed as {} + a reason) — no event methods used here.
```
Mathieu-Deharbe: Not very useful : 
```suggestion
```

src/features/network-modification-table/utils.ts
```diff
@@ -7,14 +7,27 @@
...
 export const MAX_COMPOSITE_NESTING_DEPTH = 5;
 
+interface ReferencedCompositeModifications extends NetworkModificationMetadata {
+    modificationsInfos?: NetworkModificationMetadata[];
+}
+
+export interface ReferenceModificationInfos extends NetworkModificationMetadata {
+    referenceId?: UUID;
+    referenceType?: string;
+    referenceInfos?: NetworkModificationMetadata;
```
Mathieu-Deharbe: It seems to me that `referenceInfos` will always be a `ComposedModificationMetadata` will it be not ?

So this should work and be more precise ? : 

```suggestion
    referenceInfos?: ComposedModificationMetadata;
```

src/utils/types/network-modification-metadata.ts
```diff
@@ -23,4 +23,5 @@ export interface ComposedModificationMetadata extends NetworkModificationMetadat
     subModifications: ComposedModificationMetadata[];
     maxDepth?: number;
     name?: string;
+    isSharedChild?: boolean;
```
Mathieu-Deharbe: This make it sound like this is a child which is shared. While it is the opposite : it is the child from a shared.

Maybe ? : 
```suggestion
    childFromShared?: boolean;
```

src/features/network-modification-table/renderers/reference-link-cell.tsx
```diff
@@ -0,0 +1,104 @@
...
+            setIsLoading(true);
+            // Resolve the referenceId (the GridExplore element id) and the GridExplore base URL
+            // (from the apps metadata) in parallel, then copy "<exploreUrl>/elements/<referenceId>".
+            // referenceId is resolved from the detail because the list/metadata path may not carry it.
+            Promise.all([
+                fetchNetworkModification(data.uuid).then((res) => res.json()) as Promise<ReferenceModificationInfos>,
+                fetchAppsMetadata(),
+            ])
+                .then(([detail, metadata]) => {
+                    const referenceId = detail?.referenceId;
```
Mathieu-Deharbe: ?  : 
```suggestion
                .then(([referenceInfos, metadata]) => {
                    const referenceId = referenceInfos?.referenceId;
```

src/features/network-modification-table/renderers/reference-link-cell.tsx
```diff
@@ -0,0 +1,104 @@
...
+                    const exploreUrl = metadata?.find(
+                        (app: Metadata): app is ExploreMetadata => app.name === 'Explore'
```
Mathieu-Deharbe: You could use 
```
export function isExploreMetadata(metadata: Metadata): metadata is ExploreMetadata {
    return metadata.name === 'Explore';
}
```
from `appMetadata.ts`.

src/features/network-modification-table/renderers/reference-link-cell.tsx
```diff
@@ -0,0 +1,104 @@
...
+                    )?.url;
+                    const link = `${exploreUrl}/elements/${referenceId}`;
+                    return navigator.clipboard.writeText(link).then(() => snackInfo({ headerId: 'linkCopied' }));
+                })
+                .catch((error) => snackWithFallback(snackError, error, { headerId: 'linkCopyError' }))
```
Mathieu-Deharbe, line 61: `linkCopyError` doesn't exist anywhere in the loc.

src/features/network-modification-table/renderers/reference-link-cell.tsx
```diff
@@ -0,0 +1,104 @@
...
+                </span>
+            </Tooltip>
+            <Menu
+                anchorEl={anchorEl}
+                open={open}
+                onClose={handleMenuClose}
+                slotProps={{
+                    paper: {
+                        sx: {
+                            minWidth: 0,
```
Mathieu-Deharbe: Why 0 here ?

src/features/network-modification-table/renderers/reference-link-cell.tsx
```diff
@@ -0,0 +1,104 @@
...
+                        dense: true,
+                        sx: { py: 0.25 },
+                    },
+                }}
+            >
+                <CustomMenuItem onClick={handleCopyLink} sx={{ minHeight: 0, py: 0.25, px: 1 }}>
+                    <ListItemIcon sx={{ minWidth: 24 }}>
+                        <LinkRoundedIcon sx={{ fontSize: 18, transform: 'rotate(-50deg)' }} />
+                    </ListItemIcon>
+                    <ListItemText slotProps={{ primary: { variant: 'caption' } }}>
```
Mathieu-Deharbe: What is `slotProps={{ primary: { variant: 'caption' } }}` for ?

I have never seen it anywhere else and I don't see any difference if I remove it.

src/features/network-modification-table/columns-definition.tsx
```diff
@@ -36,6 +36,10 @@ export const BASE_MODIFICATION_TABLE_COLUMNS = {
         id: 'switch',
         autoExtensible: false,
     },
+    SHARE: {
+        id: 'shared',
```
Mathieu-Deharbe: Choose between SHARE and shared.

I think shared makes more sense. This is a shared composite (from somewhere else)

.
```suggestion
    SHARED: {
        id: 'shared',
```

src/features/network-modification-table/utils.ts
```diff
@@ -244,27 +289,60 @@ export function fetchSubModificationsForExpandedRows(
...
+        fetchNetworkModification(id as UUID)
+            .then((res) => res.json())
+            .then((detail: ReferenceModificationInfos) => {
+                const children = extractReferenceChildren(detail).filter((m) => !m.stashed);
+                // Tag every direct child of a reference so cells (switch, root-network chip, name…)
+                // keep them non-clickable, just like the reference row itself.
+                const liveModifications = formatToComposedModification(children).map((m) => ({
+                    ...m,
+                    isSharedChild: true,
+                }));
```
Mathieu-Deharbe: Why do you want to make the children unclickable ?

I think it would be great for them to be editable immediately and modified everywhere. Ok we don't have the notifications yet but apart from this it should be fine (and great).

src/features/network-modification-table/utils.ts
```diff
@@ -244,27 +289,60 @@ export function fetchSubModificationsForExpandedRows(
...
+    const referenceUuidsToFetch = expandedIds.filter((id) => {
+        const mod = findModificationInTree(id, mods);
+        return isSharedModification(mod) && (force || mod?.subModifications.length === 0);
+    });
+
+    referenceUuidsToFetch.forEach((id) => {
+        fetchNetworkModification(id as UUID)
+            .then((res) => res.json())
+            .then((detail: ReferenceModificationInfos) => {
+                const children = extractReferenceChildren(detail).filter((m) => !m.stashed);
```
Mathieu-Deharbe: I don't really like the fact that we use the `referenceInfos` in order to display the content of the reference modifications.

Don't you think that we could use the referenceId of the modification reference (which is a composite uuid) and use exactly the same system than for the regular composites above ?

src/features/network-modification-table/utils.ts
```diff
@@ -26,6 +39,38 @@ export function isSharedModification(modification: ComposedModificationMetadata
...
+        ...child,
+        messageType: child.messageType ?? child.type,
+        messageValues: child.messageValues ?? '{}',
+    };
+}
+
+/**
+ * Extracts the children carried by a reference's detail payload.
+ *  - referenceInfos is a composite        -> its modificationsInfos
+ *  - referenceInfos is itself a reference -> that reference as a single (re-expandable) child
```
Mathieu-Deharbe: How could this happen ? Can we really create a reference pointing to a reference ?

src/features/network-modification-table/utils.ts
```diff
@@ -244,27 +289,60 @@ export function fetchSubModificationsForExpandedRows(
...
+                const liveModifications = formatToComposedModification(children).map((m) => ({
+                    ...m,
+                    isSharedChild: true,
+                }));
+                setMods((prev) => {
+                    const existingMod = findModificationInTree(id, prev);
+                    const mergedSubs = mergeSubModificationsIntoTree(
+                        liveModifications,
+                        existingMod?.subModifications ?? []
+                    );
```
Mathieu-Deharbe: `id` here is the id of a modification reference right ?

Then how could it possibly have subModifications ?

Actually I think 
```
const existingMod = findModificationInTree(id, prev);
                    const mergedSubs = mergeSubModificationsIntoTree(
                        liveModifications,
                        existingMod?.subModifications ?? []
                    );
```
are not useful. You could just : 
`return updateSubModificationsOfACompositeInTree(id, liveModifications, prev);`

src/features/network-modification-table/renderers/reference-link-cell.tsx
souissimai: ok

src/features/network-modification-table/utils.ts
souissimai: right

src/features/network-modification-table/utils.ts
souissimai: If fetchNetworkModification(id) already returns referenceInfos with the full nested content in one call, doing a second call to getNetworkModificationsFromComposite(referenceId) just to get the same data is redundant.

src/features/network-modification-table/utils.ts
souissimai: We will manage editing child modifications in a separate ticket. That's why I blocked users from editing a modification that is a child of another reference.

src/features/network-modification-table/renderers/reference-link-cell.tsx
souissimai: to check with stephane

src/features/network-modification-table/utils.ts
Mathieu-Deharbe: Thibault told me that there is no reason to prevent the network modification to be clickable/editable if this doesn't require extra work.

Anyway I don't know what changed but now they seems to be clickable/editable so this is fine.

src/features/network-modification-table/utils.ts
Mathieu-Deharbe: Indeed there is not reason to use fetchNetworkModification with referenceInfos AND getNetworkModificationsFromComposite. I just doubt that there is a point in populating `referenceInfos` in the back. I think that using `getNetworkModificationsFromComposite` for both composites and reference modifications would be more flexible.

But I am not sure sure so I won't insist.

src/features/network-modification-table/renderers/cell-renderers.tsx
```diff
@@ -74,6 +76,14 @@ export function DescriptionCellRenderer({ row, table }: CCtx) {
     );
 }
 
+export function ReferenceCellRenderer({ row, table }: CCtx) {
+    const { meta } = table.options;
+
+    // A reference to a shared-modification row shows a "copy link" action
+    if (isSharedModification(row.original)) {
+        return <ReferenceLinkCell data={row.original} disabled={meta?.status.isDisabled} />;
+    }
```
Mathieu-Deharbe: Agreed with coderabbit https://github.com/gridsuite/commons-ui/pull/1201#discussion_r3586396984. This is better to return null just in case : 

```suggestion
    if (isSharedModification(row.original)) {
        return <ReferenceLinkCell data={row.original} disabled={meta?.status.isDisabled} />;
    } else {
        return null;
    }
```

src/features/network-modification-table/renderers/name-cell.tsx
```diff
@@ -239,7 +239,7 @@ export function NameCell({ row, table, onChange }: Readonly<NameCellProps>) {
                 )}
                 <Box sx={createNameCellLabelBoxSx(row.getIsExpanded(), depth)}>
                     {/* Edit mode — composite only */}
-                    {isComposite && isEditing ? (
+                    {row.getCanExpand() && isEditing ? (
                         <Box
```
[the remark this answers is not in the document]
Mathieu-Deharbe: No : the reference modifications will be editable soon.

src/features/network-modification-table/renderers/reference-link-cell.tsx
```diff
@@ -0,0 +1,98 @@
...
+            // referenceId is resolved from the detail because the list/metadata path may not carry it.
+            Promise.all([
+                fetchNetworkModification(data.uuid).then((res) => res.json()) as Promise<ReferenceModificationInfos>,
+                fetchAppsMetadata(),
+            ])
+                .then(([referenceInfos, metadata]) => {
+                    const referenceId = referenceInfos?.referenceId;
+                    const exploreUrl = metadata?.find(isExploreMetadata)?.url;
+                    const link = `${exploreUrl}/elements/${referenceId}`;
+                    return navigator.clipboard.writeText(link).then(() => snackInfo({ headerId: 'linkCopied' }));
```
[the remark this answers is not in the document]
Mathieu-Deharbe: Agreed : it might be better to throw an error when referenceId or exploreUrl are undefined.

src/features/network-modification-table/row/modification-row.tsx
```diff
@@ -51,8 +51,14 @@ export function ModificationRow({
...
     );
 
+    const isChildOfReference = row.original.childFromShared === true;
+
     return (
-        <Draggable draggableId={row.id} index={virtualRow.index} isDragDisabled={isRowDragDisabled}>
+        <Draggable
+            draggableId={row.id}
+            index={virtualRow.index}
+            isDragDisabled={isRowDragDisabled || isChildOfReference}
```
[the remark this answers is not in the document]
Mathieu-Deharbe: Their drag and drop is disabled through an other way, somewhere else.

src/features/network-modification-table/renderers/reference-link-cell.tsx
Mathieu-Deharbe: Still missing : 
image

Mathieu-Deharbe (changes requested):
The only requested change is the error display : 

Image

but there are still some minor remarks left.

src/features/network-modification-table/renderers/cell-renderers.tsx
souissimai: Unnecessary 'else' after 'return'.eslint[no-else-return](https://eslint.org/docs/latest/rules/no-else-return) got this when adding the else

src/features/network-modification-table/utils.ts
souissimai: done

src/features/network-modification-table/utils.ts
souissimai: discussed

src/features/network-modification-table/utils.ts
souissimai: now they are editable

souissimai:
> The only requested change is the error display :
> Image
> 
> but there are still some minor remarks left.

add it

src/features/network-modification-table/renderers/reference-link-cell.tsx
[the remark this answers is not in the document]
souissimai: ok

src/features/network-modification-table/renderers/cell-renderers.tsx
Mathieu-Deharbe: Indeed the else is unnecessary because there is a return inside the if. But if it annoys you, you can just return null after the if.

src/translations/en/modificationEn.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Copyright (c) 2026, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+export const modificationEn = {
```
Mathieu-Deharbe: Maybe you should instead use `networkModificationsEn`.

src/translations/fr/modificationFr.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Copyright (c) 2026, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+export const modificationFr = {
```
Mathieu-Deharbe: Same here : I think it could be inside `networkModificationsFr`.

souissimai: oh ok , did not see that the file already exists sorry

Mathieu-Deharbe (approved):

src/features/network-modification-table/renderers/name-cell.tsx
```diff
@@ -239,7 +239,7 @@ export function NameCell({ row, table, onChange }: Readonly<NameCellProps>) {
                 )}
                 <Box sx={createNameCellLabelBoxSx(row.getIsExpanded(), depth)}>
                     {/* Edit mode — composite only */}
```
flomillot: Comment to update ?

src/features/network-modification-table/utils.ts
```diff
@@ -7,11 +7,21 @@
...
+interface ReferencedCompositeModifications extends NetworkModificationMetadata {
+    modificationsInfos?: NetworkModificationMetadata[];
+}
+
+export interface ReferenceModificationInfos extends NetworkModificationMetadata {
+    referenceId?: UUID;
+    referenceType?: string;
+    referenceInfos?: ComposedModificationMetadata;
+}
+
```
flomillot: It is not in the right folder for me 
Should be in  src/utils/types/network-modification-metadata.ts

src/features/network-modification-table/renderers/reference-link-cell.tsx
```diff
@@ -0,0 +1,105 @@
...
+                })
+                .catch((error) => snackWithFallback(snackError, error, { headerId: 'linkCopyError' }))
+                .finally(() => setIsLoading(false));
+        },
+        [data.uuid, snackInfo, snackError]
+    );
+
+    return (
+        <>
+            <Tooltip title={<FormattedMessage id="importComposites.shared" />} arrow enterDelay={250}>
```
flomillot: Use `CustomTooltip`

src/features/network-modification-table/renderers/root-network-chip-cell.tsx
```diff
@@ -83,6 +84,8 @@ export function RootNetworkChipCell(props: RootNetworkChipCellProps) {
     const { snackError } = useSnackMessage();
     const modificationUuid = data.uuid;
 
+    const isSharedModification = data.type === ModificationType.MODIFICATION_REFERENCE || data.childFromShared;
```
flomillot: Why having this ? 
There is already https://github.com/gridsuite/commons-ui/blob/6636af7d3a7624e8d822fd5a2aa43f3d8f19d8a9/src/features/network-modification-table/utils.ts#L35
Which should also include `childFromShared` no ?

src/features/network-modification-table/network-modifications-table.tsx
```diff
@@ -238,8 +239,8 @@ export function NetworkModificationsTable({
         getCoreRowModel: getCoreRowModel(),
         getExpandedRowModel: getExpandedRowModel(),
         getSubRows: (row) => row.subModifications,
-        getRowId: (row) => row.uuid,
-        getRowCanExpand: (row) => isCompositeModification(row.original),
+        getRowId: (row, index, parent) => (parent ? `${parent.id}.${row.uuid}` : row.uuid),
+        getRowCanExpand: (row) => isCompositeModification(row.original) || isSharedModification(row.original),
```
flomillot: But it only works because shared modification ARE composite modification, so it should not be modified

src/features/network-modification-table/network-modifications-table.tsx
```diff
@@ -238,8 +239,8 @@ export function NetworkModificationsTable({
         getCoreRowModel: getCoreRowModel(),
         getExpandedRowModel: getExpandedRowModel(),
         getSubRows: (row) => row.subModifications,
-        getRowId: (row) => row.uuid,
-        getRowCanExpand: (row) => isCompositeModification(row.original),
+        getRowId: (row, index, parent) => (parent ? `${parent.id}.${row.uuid}` : row.uuid),
```
flomillot: It only resolve partially the problem no ?  What if we use two times the same parent ?

src/features/network-modification-table/row/modification-row.tsx
```diff
@@ -51,8 +51,14 @@ export function ModificationRow({
         [handleCellClick, row.original]
     );
 
+    const isChildOfReference = row.original.childFromShared === true;
```
flomillot: Why dont you use `isTargetChildOfReference` ?

src/features/network-modification-table/use-modifications-drag-and-drop.tsx
```diff
@@ -83,6 +83,13 @@ function getContainerShadow(forbidden: boolean, isMovingDown: boolean) {
     return isMovingDown ? DROP_INDICATOR_BOTTOM : DROP_INDICATOR_TOP;
 }
 
+function isTargetChildOfReference(targetRow: Row<ComposedModificationMetadata>): boolean {
```
flomillot: Move it in utils ?

src/features/network-modification-table/use-modifications-drag-and-drop.tsx
```diff
@@ -124,6 +134,9 @@ export const useModificationsDragAndDrop = ({
                     )
                 );
             }
+            if (isTargetChildOfReference(targetRow)) {
+                return true;
+            }
```
flomillot: Move it above, then no need to compute target depth if not necessary

src/translations/en/networkModificationsEn.ts
```diff
@@ -344,4 +344,7 @@ export const networkModificationsEn = {
     StructureTab: 'Structure',
     AdditionalInformationTab: 'Additional information',
     StateEstimationTab: 'State estimation',
+    copyLink: 'Copy link',
+    linkCopied: 'Link copied',
```
flomillot, line 315: Need to remove from GridExplore then
https://github.com/gridsuite/gridexplore-app/blob/ddedfc6e070fb57f278b29ea3c0c2a871ed1522d/src/translations/en.json#L50-L51

src/features/network-modification-table/utils.ts
```diff
@@ -244,27 +275,52 @@ export function fetchSubModificationsForExpandedRows(
...
-                );
-                return updateSubModificationsOfACompositeInTree(uuid, mergedSubs, tree);
-            }, prev)
-        );
+    const referenceUuidsToFetch = expandedIds.filter((id) => {
+        const mod = findModificationInTree(id, mods);
+        return isSharedModification(mod) && (force || mod?.subModifications.length === 0);
+    });
+
+    referenceUuidsToFetch.forEach((id) => {
```
flomillot: At least, do asynchronis calls

src/features/network-modification-table/network-modifications-table.tsx
souissimai: we can expand if it's composite or shared , if we remove this condition refernce modif are not expandable

src/features/network-modification-table/renderers/root-network-chip-cell.tsx
souissimai: We should also prevent editing the applicability of shared children. If isShared is false for a child, its applicability should not be editable. Only the parent, which has the referenceModification type, should be editable.

src/features/network-modification-table/row/modification-row.tsx
souissimai: done

src/features/network-modification-table/use-modifications-drag-and-drop.tsx
souissimai: done

src/translations/en/networkModificationsEn.ts
souissimai: we can do later the clean ok TODO

src/features/network-modification-table/use-modifications-drag-and-drop.tsx
souissimai: ,we need temporary to block movve  for child

src/features/network-modification-table/renderers/reference-link-cell.tsx
souissimai: ok

src/features/network-modification-table/utils.ts
```diff
@@ -238,33 +273,61 @@ export function moveSubModificationInTree(
...
-export function fetchSubModificationsForExpandedRows(
+export async function fetchSubModificationsForExpandedRows(
     expandedIds: string[],
     mods: ComposedModificationMetadata[],
     setMods: Dispatch<SetStateAction<ComposedModificationMetadata[]>>,
     force = false
-): void {
-    const uuidsToFetch = expandedIds.filter((id) => {
+): Promise<void> {
+    const compositeUuidsToFetch = expandedIds.filter((id) => {
```
flomillot: You changed the ID of getRowId, so it can't work with nested composited if you dont change everywhere the ID is used.
Ex :

Image

src/features/network-modification-table/utils.ts
```diff
@@ -238,33 +273,61 @@ export function moveSubModificationInTree(
...
+
+    const referenceUuidsToFetch = expandedIds.filter((id) => {
+        const mod = findModificationInTree(id, mods);
+        return isSharedModification(mod) && (force || mod?.subModifications.length === 0);
     });
+
+    await Promise.all(
+        referenceUuidsToFetch.map(async (id) => {
+            try {
+                const res = await fetchNetworkModification(id as UUID);
```
flomillot: Same, log the ID with nested composite, you'll see the problem

src/features/network-modification-table/use-modifications-drag-and-drop.tsx
flomillot: Yeah which is not always the case right now, because the other test is before, right ?

src/features/network-modification-table/utils.ts
souissimai: done

src/features/network-modification-table/use-modifications-drag-and-drop.tsx
flomillot: Plus, it does not include references themself. So you can drop inside an expanded reference.

flomillot (changes requested):

src/features/network-modification-table/renderers/root-network-chip-cell.tsx
flomillot: Okay, we need both. 
But then we should have a different name because it does not represent the same thing. 
```suggestion
    const isSharedModificationOrInsideOne = data.type === ModificationType.MODIFICATION_REFERENCE || data.childFromShared;
```
Or something like that.

src/features/network-modification-table/network-modifications-table.tsx
flomillot: If it's not possible to know if a shared modification is a composite modification or not, then there is a problem in the DTO.

src/translations/en/networkModificationsEn.ts
flomillot: https://github.com/gridsuite/gridexplore-app/pull/934

src/features/network-modification-table/renderers/root-network-chip-cell.tsx
Mathieu-Deharbe: Agreed and done here : https://github.com/gridsuite/commons-ui/pull/1201/changes/44c8eeafdf4faf9fa8faa9726d381484259ac892

Except that I changed the naming. Cf next discussion.

src/features/network-modification-table/network-modifications-table.tsx
Mathieu-Deharbe: I think the naming created a misunderstanding : 
- a SHARED modification is a composite in gridexplore and remains there
- when we import it in gridstudy and SHARE it, it creates a REFERENCE modification pointing to the SHARED modification in gridexplore
Therefore there are no shared modifications in gridstudy but in a sense they all share the same mother data (from gridexplore) and for the users we write those as "shared". But as programmers we shouldn't call those "shared".

Because of this I renamed those functions : https://github.com/gridsuite/commons-ui/pull/1201/changes/44c8eeafdf4faf9fa8faa9726d381484259ac892

About the fact that all the shared modifications HAVE TO BE  composites from gridexplore : this is the case for now, but this has not been formalised and confirmed. We have to discuss this in a meeting, with at least Slimane and Thibault  in order to clarify and avoid future confusions.

src/features/network-modification-table/network-modifications-table.tsx
Mathieu-Deharbe: I don't think you can. Because when you import a composite modification as a reference modification (pointing to a shared from gridexplore) a new reference modification with its own id is created.

Therefore even if you have several reference modifications pointing to the same composite modification in gridexplore, the `row.uuid` (belonging to the netmods inside the composite) will be identical but the `parent.id` (belonging to the newly cerated reference modification) will be different.

But if I am missing something and you reproduce a bug about that I will look into it.

src/features/network-modification-table/use-modifications-drag-and-drop.tsx
Mathieu-Deharbe: No you can't drag and drop inside a reference. I guess references are tagged as childFromReference but I can't find when. Anyway it works.

And it looks like she moved the code like you requested.

src/translations/en/networkModificationsEn.ts
Mathieu-Deharbe: Approved.

src/features/network-modification-table/utils.ts
Mathieu-Deharbe: It looks like she did the work here : https://github.com/gridsuite/commons-ui/pull/1201/changes/18a1dcea44dd695d8c65508ba1d921c009514833
and here : https://github.com/gridsuite/commons-ui/pull/1201/changes/4e927a731ff76e079cb17604b382f3502b00aecd

A bit complicated, especially the drag and drop. And will be hard to understand without prior knowledge. 

I don't really have a better idea right now. But I think we should alter getRowId only when a reference modification is concerned. Those are the only cases when the same uuid may appear several times. And drag and drop between reference modifications will be very complicated anyway.

Composite modification shouldn't be changed. Thay have unique uuids and their content always have unique uuids.

Therefore I would only update the getRowId of what is inside a reference modification and only add its reference modification "mother" recursively. (ie if a netmod is inside a composite which is inside a reference modification, its id would be : referenceModificationUuid.networkModificationUuid

What do you think ? @flomillot @SlimaneAmar @souissimai 

Bonus note : maybe we could decorrelate getRowId from the modification uuid. uuid would be a separate data and not used as unique getRowId.

src/features/network-modification-table/network-modifications-table.tsx
flomillot: > About the fact that all the shared modifications HAVE TO BE composites from gridexplore : this is the case for now, but this has not been formalised and confirmed. We have to discuss this in a meeting, with at least Slimane and Thibault in order to clarify and avoid future confusions.

Exactly, that's why we shouldn't include `isShareModification` in this condition.
 `isCompositeModification` should be enough and return true if a reference modification link to a composite modification. If the DTO doesn't have the information we should add it.

src/features/network-modification-table/network-modifications-table.tsx
flomillot: Yeah I'm not sure, but as discussed, let's just use random UID.

src/features/network-modification-table/utils.ts
flomillot: As discussed we should just use random UUID

src/features/network-modification-table/use-modifications-drag-and-drop.tsx
flomillot: Ah yes it works my bad.

src/features/network-modification-table/network-modifications-table.tsx
flomillot: Add TODO as discussed

souissimai:
> I see 3 different way to access the ID of a row :
> 
> row.id row.rowKey row.orignal.rowKey
> 
> I dont get it, can you explain ?
> 
> Also you need to add the rowId for every row, not only ComposedModification, no ?

On the 3 ways to access the id: they're the same value (rowKey), read at 3 different points depending on what you're holding:

row.rowKey , only inside getRowId/getSubRows (in network-modifications-table.tsx). TanStack hands you the raw data item there, not yet wrapped, so there's no .original step.
row.original.rowKey , everywhere else you have a TanStack Row<T> (drag-and-drop, isDropForbidden...). .original unwraps it to reach the raw ComposedModificationMetadata.
row.id, TanStack's own cached copy on the Row wrapper, computed once via getRowId. Same value as row.original.rowKey, just read straight off the wrapper instead of going through .original.

On adding a rowId to every row, not just ComposedModificationMetadata: not needed elsewhere , rowKey exists specifically because in this tree, the same modification can legitimately be rendered twice at once (live in one branch, and again through a MODIFICATION_REFERENCE), which makes uuid unsafe as getRowId. Any other table whose rows never duplicate can keep using their real id directly; adding a synthetic key there would solve a problem they don't have.

To be clear: the mechanism itself hasn't changed , rows are now identified by the random rowKey, not by the modification uuid. What I fixed here is just internal consistency around it (typing ..) not a change of approach.

flomillot:
> > I see 3 different way to access the ID of a row :
> > row.id row.rowKey row.orignal.rowKey
> > I dont get it, can you explain ?
> > Also you need to add the rowId for every row, not only ComposedModification, no ?
> 
> On the 3 ways to access the id: they're the same value (rowKey), read at 3 different points depending on what you're holding:
> 
> row.rowKey , only inside getRowId/getSubRows (in network-modifications-table.tsx). TanStack hands you the raw data item there, not yet wrapped, so there's no .original step. row.original.rowKey , everywhere else you have a TanStack Row (drag-and-drop, isDropForbidden...). .original unwraps it to reach the raw ComposedModificationMetadata. row.id, TanStack's own cached copy on the Row wrapper, computed once via getRowId. Same value as row.original.rowKey, just read straight off the wrapper instead of going through .original.
> 
> On adding a rowId to every row, not just ComposedModificationMetadata: not needed elsewhere , rowKey exists specifically because in this tree, the same modification can legitimately be rendered twice at once (live in one branch, and again through a MODIFICATION_REFERENCE), which makes uuid unsafe as getRowId. Any other table whose rows never duplicate can keep using their real id directly; adding a synthetic key there would solve a problem they don't have.
> 
> To be clear: the mechanism itself hasn't changed , rows are now identified by the random rowKey, not by the modification uuid. What I fixed here is just internal consistency around it (typing ..) not a change of approach.

Thank you for the answer. 

For `row.rowKey`, I would use the same naming as in the documentation : `originalRow.rowKey`
https://tanstack.com/table/v8/docs/guide/rows#row-ids
Then we directly see it's not the same object than `row` used elsewhere (for 2 different objects, better to have 2 different names).

If row.id and row.original.rowKey are the same, why using both ? Would be better to chose only one, then we know it's the same value we're talking about.

And for the last part actually I was talking about this table only.
I though `formatToComposedModification` was used only for composite modification, but it's called actually for every modification / row. 
So yes, every original row contains `rowKey` in this table, my bad.

Also there is a problem with the type : 
https://github.com/gridsuite/commons-ui/blob/de4246a570278749f181b090b756e5f83ebc7fbc/src/features/network-modification-table/utils.ts#L344\
ReferenceModificationInfos is the type of the backend result, However it does not contain rowKey, Which makes the TypeScript wrong. 
You should either create a new type or make rowKey optional.

PR #1215, opened 2026-06-29
Add details for errors in the security analysis parameters contingency count

FranckLecuyer:


src/features/parameters/common/contingency-table/contingency-table.tsx
```diff
@@ -61,12 +77,80 @@ function ContingencyTableWithApiRef(
...
+                total += countByContingencyList.nbContingencies;
+
+                for (let j = 0; j < Object.entries(countByContingencyList.notFoundElements).length; j++) {
+                    const [contingencyName, contingencyEquipments] = Object.entries(
+                        countByContingencyList.notFoundElements
+                    )[j];
+
+                    if (
+                        contingencyEquipments !== undefined &&
+                        (contingencyEquipments as unknown as string[]).length > 0
```
TheMaskedTurtle: **[blocker] `Set<string>` typed but read as an array.** `notFoundElements` is declared `Record<string, Set<string>>` (types.ts), but here the deciding condition uses `(contingencyEquipments as unknown as string[]).length`. A `Set` has no `.length` — only `.size` — and this data comes from `fetchContingencyCount` resolving backend JSON, which cannot carry a real `Set`.

So either:
- the runtime value is actually an array → change the type to `Record<string, string[]>` and drop the cast; or
- it really is a `Set` → `.length` is always `undefined`, `undefined > 0` is always `false`, and **no invalid contingency is ever detected** (the feature silently no-ops).

The `as unknown as string[]` double cast is what hides this. Please align the declared type with the real payload and remove the cast so the compiler validates the check.

src/features/parameters/common/contingency-table/contingency-table.tsx
```diff
@@ -61,12 +77,80 @@ function ContingencyTableWithApiRef(
...
+            contingencyListsInfos.forEach((info) => {
+                info.contingencyLists.forEach((idName) => {
+                    namesById[idName.id] = idName.name;
+                });
+            });
+            const enriched: Record<string, ContingencyCountByContingencyList> = {};
+            for (let i = 0; i < Object.entries(contingencyCount.countByContingencyList).length; i++) {
+                const [contingencyListUuid] = Object.entries(contingencyCount.countByContingencyList)[i];
+                const listName = namesById[contingencyListUuid as UUID];
+                if (listName !== undefined) {
```
TheMaskedTurtle: **[major] Silent undercount when a list name is unresolved.** When `listName === undefined`, the entry is dropped from `enriched`, so contingencies from any list whose name can't be resolved are later excluded from `total` in `analyzeContingencies` — the displayed count can be wrong. Please keep unresolved lists in the total (e.g. fall back to the UUID) or justify dropping them.

src/features/parameters/common/contingency-table/contingency-table.tsx
```diff
@@ -61,12 +77,80 @@ function ContingencyTableWithApiRef(
...
+                        countByContingencyList.notFoundElements
+                    )[j];
+
+                    if (
+                        contingencyEquipments !== undefined &&
+                        (contingencyEquipments as unknown as string[]).length > 0
+                    ) {
+                        return {
+                            success: false,
+                            contingencyListName: contingencyListName.includes('_')
```
TheMaskedTurtle: **[major] Fragile string-key round-trip.** `enrichContingencyCount` encodes the name into the map key as `` `${contingencyListUuid}_${listName}` ``, and here it's parsed back out with `slice(indexOf('_') + 1)`. Encoding structured data into a string key only to split it apart again is fragile. Prefer carrying a structured value (e.g. `{ listName, count }`) so no decode step is needed.

src/features/parameters/common/contingency-table/contingency-table.tsx
```diff
@@ -61,12 +77,80 @@ function ContingencyTableWithApiRef(
...
+            contingencyListsInfos: ContingencyListsInfosEnriched[]
+        ): ContingencyCountEnriched {
+            const namesById: Record<UUID, string | undefined> = {};
+            contingencyListsInfos.forEach((info) => {
+                info.contingencyLists.forEach((idName) => {
+                    namesById[idName.id] = idName.name;
+                });
+            });
+            const enriched: Record<string, ContingencyCountByContingencyList> = {};
+            for (let i = 0; i < Object.entries(contingencyCount.countByContingencyList).length; i++) {
```
TheMaskedTurtle: **[minor] `Object.entries()` recomputed and re-indexed each iteration.** `for (let i = 0; i < Object.entries(x).length; i++)` then indexing `Object.entries(x)[i]` rebuilds the entries array on every iteration — O(n²) allocations for plain iteration. Same pattern at L121 and L128. Use `for (const [k, v] of Object.entries(x))` (or `.forEach`); faster and clearer.

src/features/parameters/common/contingency-table/contingency-table.tsx
```diff
@@ -4,24 +4,38 @@
...
 import { CONTINGENCY_LISTS } from '../constants';
 import { DEFAULT_TIMEOUT_MS, IGNORE_SIGNAL } from '../../../../services';
 import { MuiStyles, snackWithFallback, ContingencyListsInfosEnriched } from '../../../../utils';
-import { DndColumn } from '../../../../components/composite/dnd-table';
+import { DndColumn } from '../../../../components';
 
 const styles = {
-    alert: { color: 'text.primary', paddingTop: 0, paddingBottom: 0 },
+    alert: { color: 'text.primary', paddingTop: 0, paddingBottom: 0, width: '100%' },
+    error: { color: 'red', paddingTop: 0, paddingBottom: 0, width: '100%' },
```
TheMaskedTurtle: **[minor] Hardcoded `color: 'red'` instead of a theme token.** The sibling `alert` style uses `color: 'text.primary'`, and `<Alert severity="error">` already provides error coloring, so this override is likely redundant. If a color is needed, use a theme token (`error.main`). Also re-check `width: '100%'` on both `alert` and `error` — it may now be redundant with the `width: '100%'` parent `Grid`.

src/features/parameters/common/contingency-table/contingency-table.tsx
```diff
@@ -175,7 +275,7 @@ function ContingencyTableWithApiRef(
     }, [intl]);
 
     return (
-        <Grid container direction="column">
+        <Grid container sx={{ width: '100%' }}>
```
TheMaskedTurtle: The `Grid`→`Grid2` switch is fine and aligned with the ongoing migration. Just confirming: the root went from `<Grid container direction="column">` to `<Grid container sx={{ width: '100%' }}>` — please verify dropping `direction="column"` is intentional and the rows still stack as before (it reads like a deliberate layout change rather than a migration requirement).

TheMaskedTurtle (commented):
## Review — contingency count error details

**Scope:** branch review against `main` (4 files, +130/−21).
**Goal:** surface *which* contingency in *which* list is invalid, instead of just a count + "N not found".

The discriminated-union refactor (`SuccessCountType | FailureCountType`) is a clear improvement over the old loose optional-field shape, and the `Grid`→`Grid2` switch is aligned with the ongoing migration. Translations are in sync across EN/FR with matching ICU placeholders, and the removed key is no longer referenced. 👍

Main concern is one likely-blocking type/runtime mismatch (the `Set` vs `.length` cast), plus a fragile helper round-trip and a couple of minor cleanups — all inline below.

**Validation:** not run. Suggest type-check + lint; removing the `as unknown as string[]` cast should let `tsc` surface the `Set`/`.length` mismatch directly.

_Generated with the commons-ui code-review skill._

src/features/parameters/common/contingency-table/contingency-table.tsx
FranckLecuyer: code block removed ...

src/features/parameters/common/contingency-table/contingency-table.tsx
FranckLecuyer: Changed

src/features/parameters/common/contingency-table/contingency-table.tsx
FranckLecuyer: Changed

src/features/parameters/common/contingency-table/contingency-table.tsx
FranckLecuyer: Changed

src/features/parameters/common/contingency-table/contingency-table.tsx
FranckLecuyer: Changed

src/features/parameters/common/contingency-table/contingency-table.tsx
FranckLecuyer: Changed

src/features/parameters/common/contingency-table/contingency-table.tsx
```diff
@@ -112,44 +108,34 @@ function ContingencyTableWithApiRef(
...
+                    notFoundElements: 0,
                 };
             }
 
-            for (let i = 0; i < Object.entries(contingencyCount.countByContingencyList).length; i++) {
-                const [contingencyListName, countByContingencyList] = Object.entries(
-                    contingencyCount.countByContingencyList
-                )[i];
+            // eslint-disable-next-line no-restricted-syntax
+            for (const [, countByList] of Object.entries(contingencyCount.countByContingencyList)) {
```
TheMaskedTurtle: Object.values() ?

src/features/parameters/common/contingency-table/contingency-table.tsx
```diff
@@ -278,15 +263,17 @@ function ContingencyTableWithApiRef(
...
-            <ParameterTableField
-                name={name}
-                columnsDefinition={columnsDefinition}
-                tableHeight={270}
-                onChange={handleOnChange}
-                isValidRow={isValidContingencyRow}
-            />
-            {showContingencyCount && renderContingencyCount()}
+        <Grid size={12}>
+            <Stack>
```
TheMaskedTurtle: You should not use a Grid child at the top-level of a component, it should be a standalone component, here it means you know that an upper component will use a Grid container, that's not ok. Same, no need to use Satck with no props. By defautl in react, components are stacked vertically.
So something like this should be enough here
```
<>
    <ParameterTableField
                    name={name}
                    columnsDefinition={columnsDefinition}
                    tableHeight={270}
                    onChange={handleOnChange}
                    isValidRow={isValidContingencyRow}
                />
    {showContingencyCount && renderContingencyCount()}
</>
```

src/features/parameters/common/contingency-table/contingency-table.tsx
```diff
@@ -61,12 +77,68 @@ function ContingencyTableWithApiRef(
...
+            });
+
+            return {
+                countByContingencyList: enriched,
+            };
+        }
+
+        function analyzeContingencies(
+            contingencyCount: ContingencyCountEnriched | null
+        ): SimulatedContingencyCountType {
```
TheMaskedTurtle: these two functions should not be defined in a react component and even less in a useEffect. So you can declare them in a utility file instead

src/features/parameters/common/contingency-table/contingency-table.tsx
```diff
@@ -104,14 +176,17 @@ function ContingencyTableWithApiRef(
             abortSignal
         )
             .then((contingencyCount) => {
-                setSimulatedContingencyCount(contingencyCount);
+                const enrichedContingencyCount: ContingencyCountEnriched = enrichContingencyCount(
+                    contingencyCount,
+                    contingencyListsInfos
+                );
+                setSimulatedContingencyCount(analyzeContingencies(enrichedContingencyCount));
```
TheMaskedTurtle: For clarity it would be nice to have only one mapping method in the util file that you call here like 
`setSimulatedContingencyCount(mapSimulatedContingencyList(contingencyCount, contingencyListsInfos))`

src/features/parameters/common/contingency-table/contingency-table.tsx
TheMaskedTurtle: And I feel like the whole logic of these 2 functions could be simplified a lot. Can't we compute the final result directly with only one iteration loop ?

src/features/parameters/common/contingency-table/contingency-table.tsx
FranckLecuyer: kept in code simplification

src/features/parameters/common/contingency-table/contingency-table.tsx
FranckLecuyer: I removed Stack (which was introduced by Grid to Grid2 migration), but I can't remove Grid (which was already present)
You suggested change does not work ...

src/features/parameters/common/contingency-table/contingency-table.tsx
FranckLecuyer: Yes, code is moved and simplified now

src/features/parameters/common/contingency-table/contingency-table.tsx
FranckLecuyer: Changed

src/features/parameters/common/contingency-table/contingency-table.tsx
TheMaskedTurtle: It does not work because the parent component is not well defined neither ^^ 
It should be
```
<Stack sx={parametersStyles.scrollableGrid}>
            <ProviderParam options={securityAnalysisMethods.formattedProviders} id="Sa" sx={{ paddingBottom: 1 }} />
            <ContingencyTable
                name={CONTINGENCY_LISTS_INFOS}
                showContingencyCount={showContingencyCount}
                fetchContingencyCount={fetchContingencyCount}
                isBuiltCurrentNode={isBuiltCurrentNode}
                ref={contingencyTableApiRef}
            />
            <Box paddingTop={4} paddingBottom={2}>
                <LineSeparator />
            </Box>
            <SecurityAnalysisParametersSelector
                params={securityAnalysisMethods.params}
                currentProvider={securityAnalysisMethods.watchProvider?.trim()}
                isDeveloperMode={isDeveloperMode}
                defaultLimitReductions={securityAnalysisMethods.defaultLimitReductions}
            />
        </Stack>
``` 
in the form, then 
```<>
            <ParameterTableField
                name={name}
                columnsDefinition={columnsDefinition}
                tableHeight={270}
                onChange={handleOnChange}
                isValidRow={isValidContingencyRow}
            />
            {showContingencyCount && renderContingencyCount()}
        </>
```
for the Contingencytable works.

We have a lot of layouts issues in our components. So I think it's good to fix them step by step when we face the issue, so I think here it is a good occasion to fix that.

TheMaskedTurtle (changes requested):
You can just treat the layout issue and then it's good to me :)

src/features/parameters/common/contingency-table/contingency-table.tsx
FranckLecuyer: :+1: 
Changed

TheMaskedTurtle (approved):

PR #1216, opened 2026-06-29
Apply the reset of InputWithPopupConfirmation on every value change

flomillot:
## Summary

`InputWithPopupConfirmation` only ran its reset callback when the confirmation popup was shown (when `shouldOpenPopup` returns `true`). Changes that must happen on **every** value change — popup or not — had no place to live and ended up duplicated in ad-hoc effects by consumers.

The callback is now applied on every value change, whether the confirmation popup is shown or not, and is renamed accordingly: `resetOnConfirmation` becomes **`onValueChange`**, called on both paths right before the value is committed.

Consumers apply all of their resets unconditionally. Resetting more than strictly necessary is harmless: when the popup does not open, the state being reset is already empty, or equal to its default.

## Fix: the reset silently wiped the selected operational limits group ids

In `LimitsPane`, the reset replaced the whole limits sub-object, dropping the selected operational limits group ids, which can be set in view mode and were not part of that literal. Nothing was visible in the UI — the autocomplete keeps its displayed text — but the submitted modification lost them, which only showed when reopening the dialog.

This already happened on `main` when disabling the edition; applying the reset on every value change would have extended it to enabling it. The sub-object is now merged instead of replaced, and the `id` prop is used, for consistency with the field names built everywhere else in the component.


src/components/ui/reactHookForm/selectInputs/InputWithPopupConfirmation.tsx
```diff
@@ -12,7 +12,8 @@ export function InputWithPopupConfirmation({
     Input,
     name,
     shouldOpenPopup, // condition to open popup confirmation
-    resetOnConfirmation, // function to reset values in your form on confirmation,
+    resetOnChange, // function applying the changes that must happen on EVERY value change (with or without confirmation)
+    resetOnConfirmation, // complementary function applying the changes that must happen only AFTER confirmation
```
TheMaskedTurtle: Maybe we should just rename `resetOnConfirmation` to `resetOnChange`. And then it's either we have pop-up confirmation and we reset after confirmation or we don't and we reset directly on every change ? What do you think ? It combines only two logic, confirmation pop-up and reset, and not 3 logics as you did here

TheMaskedTurtle: So concretely, you just need to rename and also call the resetOnConfirmation (renamed resetOnChanged) in the else branc of handleOnchange as you did

basseche (approved):

src/components/ui/reactHookForm/selectInputs/InputWithPopupConfirmation.tsx
flomillot: That's what I tought first but actually I avoid the most I can useEffects (so many bugs with them) and I preferred to make it imperative on user interaction.
It was already the case here https://github.com/gridsuite/gridstudy-app/pull/4070/changes
And I'm not sure at 100% but I think with the useEffect we had some bugs (clearing the selected file name when opening an existing tabular modification, because it was triggering the useEffect).
Also we don't need exactly the same action when the table is empty or not. And for this, I need 2 different callback. Clearing an empty table is harmless but its still unecessary operation and less flexible for future dev.
I agree it's not super intruitive, but the description in the comments is clear IMO, maybe you have better names for the callbacks ?

ghazwarhili (approved):

src/components/ui/reactHookForm/selectInputs/InputWithPopupConfirmation.tsx
flomillot: As discussed : do the decision inside only one callback

TheMaskedTurtle (approved):

PR #1217, opened 2026-06-29
Validate CsvPicker imports against a requiredColumns list

flomillot:
## Context

Since #1166, `CsvPicker` validated the imported file header with a strict equality check (`equalsArrayAnyOrder`): the file had to contain **exactly** the expected columns. This was stricter than the previous behaviour and rejected files that legitimately omitted some optional columns (or carried extra unexisting ones).

## Change

Replace the `header` prop and the strict check with a single **`requiredColumns: string[]`** prop:

- the file is accepted as soon as it contains **every column listed in `requiredColumns`**, in any order;
- columns that match none of the expected columns are **ignored** (no longer rejected);
- pass an empty array to accept any header (every column optional).

Each consumer now declares exactly which columns are mandatory, instead of relying on an all-or-nothing toggle.

## Breaking change / consumers

`header` is removed, so every consumer must pass `requiredColumns` instead:

- gridsuite/gridstudy-app#4071 — tabular, limit-sets and tap-changer imports;
- gridsuite/gridexplore-app#908 — explicit naming contingency list.


src/components/ui/csvPicker/csv-picker.tsx
```diff
@@ -62,12 +67,14 @@ export function CsvPicker<TData = unknown>({
 
     const handleUploadAccepted = useCallback(
         (results: ParseResult<TData>, acceptedFile: File) => {
+            const actualHeader = Object.keys((results.data[0] ?? {}) as Record<string, unknown>);
```
TheMaskedTurtle: Not necessary to cast

TheMaskedTurtle (approved):

basseche (approved):

ghazwarhili (approved):

PR #1219, opened 2026-06-30
Unify table CSV buttons and truncate the imported file name

flomillot:
## Change

Rework the CSV buttons of the AG Grid table component and the imported file name display.

- **Download button**: the table CSV download now uses the shared `ExportCsvButton` (download icon) instead of an outlined text button. `ExportCsvButton` gains optional `labelId`/`onClick`.
- **Reusable `CsvDownloadButton`**: extracted from `BottomTableButtons` and exported, with `withExportButton` (icon) and `variant` options. The "Generate template" button is **removed from the bottom toolbar** so each form can place it itself (contained) above the table; `getTemplateData` is dropped from `CsvProps`.
- **File name truncation**: in `CsvPicker`, the imported file name is shown on a single line, truncated with an ellipsis, with a tooltip showing the full name — a long name no longer wraps to an extra line.
- **Labels**: `Upload CSV` → `Import CSV`, `Generate CSV template` → `Generate template` (FR: « Importer un CSV », « Générer le modèle »).
- `ExplicitNamingFilterForm` is updated to render the generate-template button itself.
- Remove the buggy upload progress bar from `CsvPicker` (not wanted by UX).
- `ExplicitNamingFilterForm` now hides everything below the equipment type selector until a type is selected, instead of disabling the buttons and hiding only the table.
- Restore the row spacing lost when the root `Grid` container was replaced by a `Stack`: a nested `Grid` container only inherits the spacing CSS variables from a parent `Grid` container, so the row now declares `spacing` explicitly.

## Breaking change / consumers

`CsvProps.getTemplateData` is removed and the bottom "Generate template" button no longer renders automatically: every `CustomAgGridTable` consumer must render the button itself via the exported `CsvDownloadButton`.

- gridsuite/gridstudy-app — tabular & limit-sets forms
- gridsuite/gridexplore-app — explicit naming contingency list


src/components/ui/csvPicker/csv-picker.tsx
```diff
@@ -106,29 +106,39 @@ export function CsvPicker<TData = unknown>({
...
+                                }}
+                            >
+                                {selectedFile ? selectedFile.name : intl.formatMessage({ id: 'uploadMessage' })}
+                            </Box>
+                        </Tooltip>
+                        <Button
+                            {...getRootProps()}
+                            variant={selectedFile ? 'contained' : 'text'}
+                            disabled={disabled}
+                            sx={{ flexShrink: 0 }}
```
TheMaskedTurtle: I would recommend something like this to remain more consistent with the rest of the layout code
```suggestion
                     <Grid2 container spacing={1} alignItems="center" justifyContent="right" wrap="nowrap">
                        <Grid2>
                            <Tooltip title={selectedFile?.name}>
                                <Typography
                                    sx={{
                                        fontWeight: 'bold',
                                        overflow: 'hidden',
                                        textOverflow: 'ellipsis',
                                    }}
                                >
                                    {selectedFile ? selectedFile.name : <FormattedMessage id="uploadMessage" />}
                                </Typography>
                            </Tooltip>
                        </Grid2>
                        <Grid2 size="auto">
                            <Button
                                {...getRootProps()}
                                variant={selectedFile ? 'contained' : 'text'}
                                disabled={disabled}
                            >
                                <FormattedMessage id={label} />
                            </Button>
                        </Grid2>
                    </Grid2>
```

TheMaskedTurtle (commented):
In the structure it is ok for me, except for the remark about the layout. 

Besides, I haven't really tested it functionally and I'm not aware of UX validations. So it requires an additionnal review from your team.

src/components/composite/agGridTable/BottomTableButtons.tsx
```diff
@@ -74,25 +42,16 @@ export function BottomTableButtons({
...
                 {csvProps?.getTableData && (
                     <Grid>
                         <CsvDownloadButton
                             data={csvProps.getTableData}
                             fileName={csvProps.fileName}
-                            delimiter={getCsvDelimiter(csvProps.language)}
+                            language={csvProps.language}
                             labelId="DownloadCSV"
                             disabled={!csvProps.hasTableData}
+                            withExportButton
```
ghazwarhili, line 54: why this changes ?

src/components/ui/csvDownloader/csv-download-button.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { getCsvDelimiter } from '../../../utils';
+import { ExportCsvButton } from './export-csv-button';
+
+export interface CsvDownloadButtonProps {
+    labelId: string;
+    data: () => unknown[];
+    fileName: string;
+    language?: string;
+    // When true, render the shared ExportCsvButton (icon) instead of a plain text Button.
+    withExportButton?: boolean;
```
ghazwarhili, line 20: The prop has a default value, so it's not optional ?

src/components/ui/csvDownloader/csv-download-button.tsx
```diff
@@ -0,0 +1,55 @@
...
+    disabled?: boolean;
+    variant?: ButtonProps['variant'];
```
ghazwarhili, line 22: the same for disabled and variant

src/components/ui/csvDownloader/csv-download-button.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    if (disabled) {
+        return button;
+    }
+
+    return (
+        <CSVDownloader data={data} filename={fileName} config={{ delimiter: getCsvDelimiter(language) }}>
+            {button}
+        </CSVDownloader>
+    );
```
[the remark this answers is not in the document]
ghazwarhili: @flomillot  I agree with Rabbit's comments. Could you take a look?

src/components/ui/csvDownloader/csv-download-button.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <ExportCsvButton labelId={labelId} disabled={disabled} />
+    ) : (
+        <Button variant={variant} disabled={disabled}>
+            <FormattedMessage id={labelId} />
+        </Button>
+    );
+
+    if (disabled) {
+        return button;
+    }
```
ghazwarhili, line 48: const button = withExportButton ? (
        <ExportCsvButton labelId={labelId} disabled={disabled} />
    ) : (
        <Button variant={variant} disabled={disabled}>
            <FormattedMessage id={labelId} />
        </Button>
    );

    if (disabled) {
        return button;
    }

Why all of this? We can simplify it to make it easier to understand

src/components/ui/csvDownloader/export-csv-button.tsx
```diff
@@ -10,21 +10,23 @@ import { FormattedMessage } from 'react-intl';
 import { GetApp as GetAppIcon, Check as CheckIcon } from '@mui/icons-material';
 
 export interface ExportButtonProps {
+    labelId?: string;
```
ghazwarhili, line 13: labelId = 'MuiVirtualizedTable/exportCSV' already default value
why it is optional ?

basseche (approved):

src/components/ui/csvDownloader/csv-download-button.tsx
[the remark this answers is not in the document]
flomillot: I understand but what's the quick solution ? The UX wants this button.
We're stuck with react-papaparse, I asked which lib we should use but no answer yet so I kept it.
If you know what to do I'm opened to suggestion. But at least it works. The button is just here for the design.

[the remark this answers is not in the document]
flomillot: We can delay this topic until we know if we should delete react-papaparse and instead use this hook I guess, if it's OK for you 
https://github.com/gridsuite/commons-ui/blob/5dfdbaa92c7868f261c3677aaf25fc5a62618c5d/src/components/ui/csvDownloader/use-csv-export.ts#L21

src/components/ui/csvPicker/csv-picker.tsx
flomillot: I kept the previous components but yes better, I changed it

src/components/composite/agGridTable/BottomTableButtons.tsx
flomillot: Need to choose which button to display

src/components/ui/csvDownloader/csv-download-button.tsx
flomillot: I dont understand, its because it has a default value that its optional

flomillot: I mean I could remove the default value and yes it still works, but I think its clearer like that

src/components/ui/csvDownloader/csv-download-button.tsx
flomillot: Actually it is not coherent because disabled should be false by default, I changed it

src/components/ui/csvDownloader/csv-download-button.tsx
flomillot: How would you simplify ?

src/components/ui/csvDownloader/export-csv-button.tsx
flomillot: Same answer as above, not sure to understand

src/components/ui/csvPicker/csv-picker.tsx
```diff
@@ -106,29 +106,25 @@ export function CsvPicker<TData = unknown>({
...
-                    </>
+                {({ getRootProps }: any) => (
+                    <Grid container spacing={1} alignItems="center" justifyContent="right">
+                        <Grid>
+                            <Tooltip title={selectedFile?.name}>
+                                <Typography noWrap sx={{ fontWeight: 'bold' }}>
+                                    {selectedFile ? selectedFile.name : <FormattedMessage id="uploadMessage" />}
+                                </Typography>
+                            </Tooltip>
+                        </Grid>
```
TheMaskedTurtle:
```suggestion
                    <Grid container spacing={1} alignItems="center" justifyContent="right" wrap="nowrap">
                        <Grid>
                            <Tooltip title={selectedFile?.name}>
                                <Typography sx={{ fontWeight: 'bold', overflow: 'hidden', textOverflow: 'ellipsis' }}>
                                    {selectedFile ? selectedFile.name : <FormattedMessage id="uploadMessage" />}
                                </Typography>
                            </Tooltip>
                        </Grid>
```
I feel this works better, you should test with a file that has a long name, you'll see

src/components/composite/filter/explicitNaming/ExplicitNamingFilterForm.tsx
```diff
@@ -233,33 +233,43 @@ export function ExplicitNamingFilterForm({
...
-                        validateButtonLabel="button.changeType"
-                        sx={{ width: 400, maxWidth: '100%' }}
-                        data-testid="EquipmentTypeSelector"
+                    <CsvDownloadButton
+                        data={getTemplateData}
+                        fileName={intl.formatMessage({ id: 'filterCsvFileName' })}
+                        language={language}
+                        labelId="GenerateCSV"
+                        variant="contained"
+                        disabled={!watchEquipmentType}
```
TheMaskedTurtle: I this it's a bit weird to display the button when no equipment type is selected

TheMaskedTurtle (commented):
I'm a bit confused about the global goal. We want to harmonize download and upload CSV buttons right ? Then, why don't we have only one CsvDownloadButton and one CsvPicker (already good) ? Why we maintain different options for the CSV downloader ? I would keep only one option with the icon button and put that everywhere. 
Do we plan to migrate all the download/upload buttons then ? Because I feel like a lot of components still carry their own implementation.

src/components/ui/csvPicker/csv-picker.tsx
flomillot: Ah indeed my bad my HMR didnt work when I tested.
Btw noWrap from MUI apply `overflow: 'hidden', textOverflow: 'ellipsis',  whiteSpace: 'nowrap'`, so it's like removing  ` whiteSpace: 'nowrap'`.

With
image
Without 
image

I agree I think it's better, I change it.

src/components/composite/filter/explicitNaming/ExplicitNamingFilterForm.tsx
flomillot: It haven't been requested by the UX, but I agree.
I will hide everything under equipmentType if it's undefined.

ghazwarhili (approved):

TheMaskedTurtle (approved):
Not tested again after last changes but OK on the principle. It should have a follow-up to clarify the situation, currently we have a lot of different paths to generate a CSV from an AgGrid table it would be nice to have only one. Anyway, we could still have one button and several "hooks" to generate the CSV data, so it's a first step but that's an on-going work.

PR #1221, opened 2026-07-02
chore: catch TS declarations errors in CI

TheMaskedTurtle:
## PR Summary

TS errors can appear only during declaration files creation. Especially the TS2742 "The inferred type of SpecificType cannot be named without a reference to 'specific-module'". 

Currently these types of errors are never caught by the CI and can be shipped in production builds. This can cause TS errors in @gridsuite/commons-ui package consumers. 

Indeed, for now we either type-check and don't emit declaration files (npm run type-check) or emit declaration files and don't type check (vite-dts-plugin). Then these errors are missed in the first case, and logged but ignored in the second one. 

This adding reinforces only the type-check used in the CI so that it also generates declaration files to be able to perform a comprehensive type check. We can still run 'npm run type-check' for faster compilation check during development.


achour94 (approved):

flomillot (approved):

antoinebhs:
Maybe add entry to readme to explain what is this?
Do we need this in other apps too?

TheMaskedTurtle:
> Maybe add entry to readme to explain what is this? Do we need this in other apps too?

@antoinebhs Ok I can add in README. And no, we don't need it in apps, it's on the generation of declarations files so only for libs.

antoinebhs (approved):
🔝

PR #1229, opened 2026-07-06
[FIX] LF, SA, SENSI, SC, NETWORK, PCC, VL params dialogs crashed or not stable

thangqp:
## PR Summary

- Form should not be rendered before params loaded. There are some atomic input components may be not stable (e.g. autocomplete) when value/options is undefined.
- Bug only in explore which has not yet protect by checking loading or not. This has beeen done in study in the parameter layout, see https://github.com/gridsuite/commons-ui/blob/main/src/features/parameters/common/parameter-layout/parameter-layout.tsx#L121


antoinebhs (commented):
Why dynamic parameters don't have this?

antoinebhs (approved):
Code OK

PR #1230, opened 2026-07-07
Fix empty default loadflow params connected component

antoinebhs:
## PR Summary

Connected component is empty by default because we're missing the default option:
image


EtienneLt (approved):
Code Ok test Ok

ayolab (approved):

PR #1232, opened 2026-07-07
chore(LineModification): move form from grid-study

dbraquart:
## PR Summary

The goal is to move the LineModification form from grid-study to this library. The dialog remains in grid-study.


src/translations/fr/networkModificationsFr.ts
```diff
@@ -312,6 +312,8 @@ export const networkModificationsFr = {
     NoBusbarSectionFound: 'Aucun jeu de barres trouvé pour ce poste',
     CreateLine: 'Créer une ligne',
     LineCreationError: "Erreur lors de la création d'une ligne",
```
JacquesBorsenberger, line 314: i know that the i18n is not sorted, but it seems to have the same order in each files

achour94 (approved):
code and test ok

ayolab (approved):

PR #1242, opened 2026-07-16
Add ElementAttributes status

Meklo:


src/utils/types/types.ts
```diff
@@ -25,6 +25,7 @@ export type ElementAttributes = {
     lastModificationDate: string;
     lastModifiedBy: string; // id
     lastModifiedByLabel?: string; // enrich with user identity server
+    status: string;
```
flomillot: Not true because not always present
ex : 
https://github.com/gridsuite/gridexplore-app/blob/0b2c5f690b0213443ab0030858703362c5a9d0fe/src/hooks/useDirectoryContent.ts#L78
https://github.com/gridsuite/commons-ui/blob/27178aa5c9eee14d92cbc791c7b1bcf98ad2dbb1/src/services/explore.ts#L77
https://github.com/gridsuite/explore-server/blob/ad3a9818b6765dfef47fcbf1d252bafdc1b10846/src/main/java/org/gridsuite/explore/server/services/DirectoryService.java#L343

flomillot (changes requested):

src/utils/types/types.ts
flomillot: Also maybe include 
`export type ElementStatus = 'ACTIVE' | 'DELETING';` 
?

flomillot: Then `status?: ElementStatus`

Meklo: done

Meklo: went for an enum in the end

flomillot (approved):

PR #1244, opened 2026-07-16
Move modification by formula

anass45:
Moving the Modification by formula form, from gridstudy to commons-ui


KoloMenek (approved):

flomillot (approved):
formula/formula is a bit weird, but we already have the same for assignment/assingment, so well lets stay like that

PR #1245, opened 2026-07-17
add voltage regulation for battery creation

EtienneLt:


Mathieu-Deharbe (changes requested):
It looks like Reactive percentage is lost when closing and reopening the battery creation panel : 

Image

(it works in the generator creation)

src/features/network-modifications/battery/creation/BatteryCreationForm.tsx
```diff
@@ -60,6 +70,20 @@ export function BatteryCreationForm({
         />
     );
 
+    const voltageRegulationField = (
+        <Box>
+            <SwitchInput name={FieldConstants.VOLTAGE_REGULATION} label="VoltageRegulationText" />
+        </Box>
+    );
+
+    const voltageRegulationFields = (
```
Mathieu-Deharbe: Maybe : ?
```suggestion
    const voltageRegulationForm = (
```
Because it is a bit strange to have a variable `voltageRegulationField` and another `voltageRegulationFields`. But you could also merge them together and call the merge `voltageRegulationFields`.

src/features/network-modifications/battery/creation/BatteryCreationForm.tsx
```diff
@@ -89,6 +113,13 @@ export function BatteryCreationForm({
 
             {/* Active power control part */}
             <Grid container spacing={2} paddingTop={2}>
+                <GridItem
+                    tooltip={watchVoltageRegulation !== null ? '' : <FormattedMessage id="NoModification" />}
```
Mathieu-Deharbe: Agreed with coderabbit : 'NoModification' tooltip makes no sense  here in a creation. Maybe "No voltage regulation'  ? Which would mean that we won't add any field about it during the creation.

But more importantly : I can't display it. I don't think that watchVoltageRegulation may be null. Are you sure that there is a way to display it ? Otherwise just remove this.

Mathieu-Deharbe (approved):

ayolab (approved):

PR #1246, opened 2026-07-17
feat: add optional references field to ElementAttributes

flomillot:
Expose the `references` of an element on `ElementAttributes`, along with the `ReferenceAttributes` type and the `ReferenceType` enum.

A shared element carries the list of its uses as references. Applications need them to tell whether an element is shared, and to list the elements using it.

Required by gridsuite/gridexplore-app#895.


Meklo (approved):

src/utils/types/types.ts
```diff
@@ -10,6 +10,17 @@ import { ElementType } from './elementType';
 
 export type Input = string | number;
 
+export enum ReferenceType {
```
ghazwarhili, line 18: ReferenceType a bit too generic here 
change to a specific naming

src/utils/types/types.ts
```diff
@@ -10,6 +10,17 @@ import { ElementType } from './elementType';
...
+    STUDY_NODE = 'STUDY_NODE',
+    NETWORK_MODIFICATION = 'NETWORK_MODIFICATION',
+    DIRECTORY_ELEMENT = 'DIRECTORY_ELEMENT',
+}
+
+export type ReferenceAttributes = {
```
ghazwarhili, line 24: type or interface ?

src/utils/types/types.ts
```diff
@@ -38,6 +49,7 @@ export type ElementAttributes = {
     subtype?: string;
     // only uploading element have this field
     id?: string;
+    references?: ReferenceAttributes[];
```
ghazwarhili, line 58: the default value here is [] ? or it is optional/undefined ?

src/utils/types/types.ts
flomillot: This is the name used in the back-end
https://github.com/gridsuite/directory-server/blob/0ce0295c42d217336b7a92e28344d9dd6b221314/src/main/java/org/gridsuite/directory/server/dto/ReferenceAttributes.java#L26

src/utils/types/types.ts
flomillot: Type by coherence with existing code aside ?
image

src/utils/types/types.ts
flomillot: If it's optional, that mean we accept null or undefined values

ghazwarhili (approved):

klesaulnier (approved):

PR #1248, opened 2026-07-20
Fix for lost focus on typing on filter creation/editing

JacquesBorsenberger:
## PR Summary

Fix for lost focus on typing - ensure that all parmeters under  QueryBuilderMaterial are stable


src/components/composite/reactQueryBuilder/CustomReactQueryBuilder.tsx
```diff
@@ -61,6 +61,23 @@ const customTranslations = {
     combinators: { title: '' },
 };
 
+const dnd = { ...ReactDnD, ...ReactDndHtml5Backend };
```
thangqp, line 64: Following to the doc here: https://react-querybuilder.js.org/docs/dnd
Using directly raw react-dnd exported is still backward-compatible but IMO, it is better to update with the new usage in their documentation:
const adapter = createReactDnDAdapter({ ...ReactDnD, ...ReactDnDHTML5Backend });
This usage does not fix the bug absence of drag preview which is actually in prod and acceptance env. but may help us better understanding how react query builder adapt to a dnd library.

thangqp: I have tried to use  @atlaskit/pragmatic-drag-and-drop adapter, as in the doc, it works but the drag preview is smaller than usual.
 @dnd-kit adapter does not work at all, even the drag icon is not shown.

See with Florian, the fix for lost focus is priori than Drag Preview. I will create a Jira ticket for Drag Preview bug

thangqp (approved):

src/components/composite/reactQueryBuilder/CustomReactQueryBuilder.tsx
JacquesBorsenberger: i think it's better to enhance this part on a dedicated PR fixing the DnD

antoinebhs (approved):
code ok

PR #1250, opened 2026-07-21
Use only authorized server to make rest calls

klesaulnier:
## PR Summary

Migrate all API calls in order to use only authorized servers


src/services/dynamic-mapping.ts
```diff
@@ -6,11 +6,10 @@
  */
 import { backendFetchJson } from './utils';
 import { DynamicSimulationModelInfos, IdName } from '../utils';
-
-const PREFIX_DYNAMIC_MAPPING_SERVER_QUERIES = `${import.meta.env.VITE_API_GATEWAY}/dynamic-mapping`;
+import { PREFIX_STUDY_SERVER_QUERIES } from './study';
 
 function getDynamicMappingUrl() {
-    return `${PREFIX_DYNAMIC_MAPPING_SERVER_QUERIES}/`;
+    return `${PREFIX_STUDY_SERVER_QUERIES}/v1/dynamic-mapping/`;
```
antoinebhs: dynamic-mapping will be exposed in the gateway, it's the bff of griddyna. I think we can keep it

src/services/dynamic-security-analysis.ts
```diff
@@ -23,13 +21,13 @@ export function fetchContingencyAndFiltersLists(listIds: UUID[]): Promise<Elemen
 
     urlSearchParams.append('strictMode', 'false');
 
-    const url = `${PREFIX_DIRECTORY_SERVER_QUERIES}/v1/elements?${urlSearchParams}`;
+    const url = `${PREFIX_STUDY_SERVER_QUERIES}/v1/directory/elements?${urlSearchParams}`;
```
antoinebhs, line 24: I wonder if we want this or move those to study? it seems strange to have dynamic-security-analysis.ts only for some and other are in study historically?

src/services/userAdmin.ts
```diff
@@ -6,21 +6,20 @@
...
+    return backendFetchJson(`${PREFIX_STUDY_SERVER_QUERIES}/v1/user-admin/users/${user}/detail`, {
         method: 'get',
         headers: { 'Content-Type': 'application/json' },
     });
 }
 
 export function fetchCurrentAnnouncement(): Promise<AnnouncementDto | null> {
     console.info(`Fetching current announcement ...`);
-    const url = `${PREFIX_USER_ADMIN_QUERIES}/announcements/current`;
+    const url = `${PREFIX_STUDY_SERVER_QUERIES}/v1/user-admin/announcements/current`;
```
antoinebhs: user admin is exposed in the gateway?

PR #1261, opened 2026-07-27
feat: expose the write permission on shared modifications

flomillot:
## PR Summary

Investigation/POC: honour the WRITE permission on shared modifications inside a study's modification table.

A shared modification is a `MODIFICATION_REFERENCE` carrying a `referenceId`, which is also the `elementUuid` of
the composite in directory-server. `use-shared-modifications-permissions` resolves
`hasElementPermission(referenceId, WRITE)` per shared modification, caches the answers and invalidates them on
`UPDATE_DIRECTORY` notifications.

Two distinct sets are exposed: `readOnlySharedModificationUuids` (the shared modifications the user cannot write
into) and `lockedNestedModificationUuids` (their content). Actions on the shared modification as a whole stay
allowed — only its content is locked: cell editing, rename, description save, form opening, drag source and
target.

Requires gridsuite/network-modification-server#858: without `referenceId` in the metadata nothing locks, so that
one has to be merged first.

Limitations left as TODO in the code: one permission call per reference (`/elements/authorized` is
all-or-nothing, no batch endpoint), and a permission granted through a group is not invalidated —
user-admin-server emits no notification on group composition change.

### DIFF
--- a/src/features/network-modification-table/renderers/description-cell.tsx
+++ b/src/features/network-modification-table/renderers/description-cell.tsx
@@ -20,10 +20,12 @@ export interface DescriptionCellProps {
     studyUuid: UUID | null;
     currentNodeId?: UUID;
     isDisabled?: boolean;
+    // the dialog stays reachable, only its validation is denied
+    isSaveDisabled?: boolean;
 }
 
 export function DescriptionCell(props: DescriptionCellProps) {
-    const { data, studyUuid, currentNodeId, isDisabled = false } = props;
+    const { data, studyUuid, currentNodeId, isDisabled = false, isSaveDisabled = false } = props;
     const [isLoading, setIsLoading] = useState(false);
     const [openDescModificationDialog, setOpenDescModificationDialog] = useState(false);
 
@@ -60,6 +62,7 @@ export function DescriptionCell(props: DescriptionCellProps) {
                     description={description ?? ''}
                     onClose={handleDescDialogClose}
                     updateElement={updateModification}
+                    disabledSave={isSaveDisabled}
                 />
             )}
             <Tooltip title={description ?? <FormattedMessage id="addDescription" />} arrow enterDelay={250}>
--- a/src/features/network-modification-table/renderers/name-cell.tsx
+++ b/src/features/network-modification-table/renderers/name-cell.tsx
@@ -37,9 +37,10 @@ interface NameCellProps {
     row: Row<ComposedModificationMetadata>;
     table: Table<ComposedModificationMetadata>;
     onChange?: (modification: ComposedModificationMetadata, newValue: string) => Promise<unknown>;
+    isRenameDisabled?: boolean;
 }
 
-export function NameCell({ row, table, onChange }: Readonly<NameCellProps>) {
+export function NameCell({ row, table, onChange, isRenameDisabled = false }: Readonly<NameCellProps>) {
     const { meta } = table.options;
     const intl = useIntl();
     const theme = useTheme();
@@ -49,6 +50,7 @@ export function NameCell({ row, table, onChange }: Readonly<NameCellProps>) {
     const { depth } = row;
 
     const isComposite = isCompositeModification(row.original);
+    const isRenamable = isComposite && !isRenameDisabled;
 
     const getModificationLabel = useCallback(
         (modification: ComposedModificationMetadata, formatBold: boolean = true) => {
@@ -156,13 +158,13 @@ export function NameCell({ row, table, onChange }: Readonly<NameCellProps>) {
     // triggers composite name editing from outside the component
     useEffect(() => {
         const modificationToEditLabel = meta?.interaction.modificationToEditLabel.current;
-        if (isComposite && !isEditingRef.current && modificationToEditLabel === row.original.uuid) {
+        if (isRenamable && !isEditingRef.current && modificationToEditLabel === row.original.uuid) {
             beginEditingName(defaultCompositeName);
             if (meta) {
                 meta.interaction.modificationToEditLabel.current = null;
             }
         }
-    }, [meta, defaultCompositeName, isComposite, row.original.uuid, beginEditingName]);
+    }, [meta, defaultCompositeName, isRenamable, row.original.uuid, beginEditingName]);
 
     const handleLabelClick = useCallback(
         (e: React.MouseEvent) => {
@@ -195,7 +197,7 @@ export function NameCell({ row, table, onChange }: Readonly<NameCellProps>) {
             <DepthBox key={i} firstLevel={i === 0} displayAsFolder={isComposite && i === depthLevelCount - 1} />
         ));
     };
-    const compositeReadModeProps = isComposite
+    const renamableModeProps = isRenamable
         ? {
               ref: labelRef,
               onClick: handleLabelClick,
@@ -272,11 +274,11 @@ export function NameCell({ row, table, onChange }: Readonly<NameCellProps>) {
                         /* Read mode */
                         <CustomTooltip disableFocusListener disableTouchListener title={label}>
                             <Box
-                                {...compositeReadModeProps}
+                                {...renamableModeProps}
                                 sx={mergeSx(
                                     networkModificationTableStyles.modificationLabel,
                                     createModificationNameCellStyle(row.original.activated),
-                                    compositeReadModeProps.sx
+                                    renamableModeProps.sx
                                 )}
                             >
                                 {label}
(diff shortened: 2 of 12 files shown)


src/features/network-modification-table/renderers/name-cell.tsx
```diff
@@ -154,6 +156,7 @@ export function NameCell({ row, table, onChange }: Readonly<NameCellProps>) {
     const defaultCompositeName: string = useMemo(() => intl.formatMessage({ id: 'CompositeModification' }), [intl]);
 
     // triggers composite name editing from outside the component
+    // i.e., when a composite is being created
```
ghazwarhili, line 159: Unnecessary comment !
revert

src/features/network-modification-table/renderers/description-cell.tsx
```diff
@@ -52,6 +54,11 @@ export function DescriptionCell(props: DescriptionCellProps) {
         setOpenDescModificationDialog(true);
     }, []);
 
+    // Nothing to read and nothing to write
```
ghazwarhili: Unnecessary comment !
remove it

src/utils/constants/notificationsProvider.ts
```diff
@@ -13,6 +13,12 @@ export enum NotificationsUrlKeys {
     DIRECTORY_DELETE_STUDY = 'DIRECTORY_DELETE_STUDY',
     MONITOR = 'MONITOR',
 }
+export enum DirectoriesNotificationType {
+    DELETE_DIRECTORY = 'DELETE_DIRECTORY',
+    ADD_DIRECTORY = 'ADD_DIRECTORY',
+    UPDATE_DIRECTORY = 'UPDATE_DIRECTORY',
+}
```
ghazwarhili, line 21: This isn't the right file for this enum !

src/features/network-modification-table/renderers/description-cell.tsx
flomillot: Returning null from a component is not something that is common and should be explained. At least it's legit.

flomillot: I update the comments to make it clearer.

src/features/network-modification-table/renderers/name-cell.tsx
flomillot: I don't think it is. The only purpose of this use effect is when we create a new composite which was not mentioned.

src/utils/constants/notificationsProvider.ts
flomillot: I agree the file does not fit totally, but I didn't find an accurate file. Where is it ?

src/utils/types/network-modification-metadata.ts
```diff
@@ -17,6 +17,8 @@ export interface NetworkModificationMetadata {
     description: string;
     messageType: string;
     messageValues: string;
+    // MODIFICATION_REFERENCE only: uuid of the referenced composite modification
+    referenceId?: UUID;
```
ghazwarhili, line 21: referenceId is already declared in ReferenceModificationInfos could that cause conflicts?

ghazwarhili:
Here’s my code review of this https://github.com/gridsuite/commons-ui/pull/1313

src/utils/types/network-modification-metadata.ts
flomillot: Nope because it's the same field in the back. However ReferenceModificationInfos is bad typed because it should not be optional.

flomillot: However, I added this line before the one you mention exists, so I can remove the other one.

src/features/network-modification-table/renderers/description-cell.tsx
```diff
@@ -52,6 +54,11 @@ export function DescriptionCell(props: DescriptionCellProps) {
         setOpenDescModificationDialog(true);
     }, []);
 
+    // As the description is empty and we can't update it, we don't want to render the cell and its button
+    if (empty && isSaveDisabled) {
```
ghazwarhili, line 58: and if openDescModificationDialog = true 
(component is already mounted)

src/features/network-modification-table/use-shared-modifications-permissions.ts
```diff
@@ -0,0 +1,121 @@
...
+        if (missingIds.length === 0) {
+            // no fetch needed
+            return undefined;
+        }
+
+        getAccessibleElements(missingIds, PermissionType.WRITE)
+            .catch((error) => {
+                console.error('Failed to resolve the permissions on the shared modifications', error);
+                // Denying them all is what the cache already reports while they are unresolved
+                return [];
```
ghazwarhili: why return [] ? (should be return;)
even if the  error is about fetch failure ?  (it will be treaten like a real deny)

src/features/network-modification-table/use-shared-modifications-permissions.ts
```diff
@@ -0,0 +1,121 @@
...
+            })
+            .then((accessibleIds) => {
+                if (aborted) {
```
ghazwarhili, line 99: if (aborted || accessibleIds == null)  return;

src/features/network-modification-table/use-shared-modifications-permissions.ts
```diff
@@ -0,0 +1,121 @@
...
+    }, []);
+
+    useNotificationsListener(NotificationsUrlKeys.DIRECTORY, {
+        listenerCallbackMessage: handleDirectoryNotification,
+    });
+
+    useEffect(() => {
+        let aborted = false;
+
+        const referenceIds = getReferenceIds(modifications.filter(isReferenceModification));
```
ghazwarhili, line 77: use useMemo for referenceIds (outside the useEffect)

flomillot: Use memo is useful when there is a heavy compute or when you don't want to trigger re-render. 
It's not the case here.

src/features/network-modification-table/renderers/description-cell.tsx
flomillot: I don't understand what you want to change exactly. Can you suggest the code here ?

src/features/network-modification-table/use-shared-modifications-permissions.ts
flomillot: Actually, it was to avoid allowing permission if we don't know if the user has the rights. 
But it's not necessary anymore and it removes them from missing IDs which is wrong. 
I change it.

src/features/network-modification-table/use-shared-modifications-permissions.ts
flomillot: I switched the order of then and catch, then I don't need this.

src/features/network-modification-table/row/modification-row.tsx
```diff
@@ -28,6 +28,8 @@ interface ModificationRowProps {
     handleCellClick?: (modification: ComposedModificationMetadata) => void;
     isRowDragDisabled: boolean;
     highlightedModificationUuid: string | null;
+    // TODO temporary before GRD-5139
```
etiennehomer: no ticket number in the code

src/features/network-modification-table/use-shared-modifications-permissions.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+// TODO a permission granted through a group stays cached when the user is added to / removed from that group:
+// user-admin-server emits no notification on group membership changes, unlike directory-server on permissions.
+// Also consider deleting this hook and using Redux instead if we start using this hook at several places.
+export function useSharedModificationsPermissions(modifications: NetworkModificationMetadata[]): {
+    readOnlySharedModificationUuids: Set<UUID>;
+} {
+    const [readOnlySharedModificationUuids, setReadOnlySharedModificationUuids] = useState<Set<UUID>>(EMPTY_UUID_SET);
+    // referencedId -> has the write permission
+    const [permissionsCache, setPermissionsCache] = useState<Map<UUID, boolean>>(() => new Map());
```
etiennehomer: is it really a cache as it is updated when directory's rights change ?

src/utils/constants/notificationsProvider.ts
etiennehomer: new file in utils/types or in utils/types/types.ts

flomillot:
Closing in favour of gridsuite/commons-ui#1368, which reads the permission the server now carries on each shared modification instead of resolving it in the browser. The shared declaration of the directory notification types moved to gridsuite/commons-ui#1366, so that gridsuite/gridexplore-app#956 is not held back by this work.

PR #1262, opened 2026-07-27
add voltage regulation for battery modification

EtienneLt:

### DIFF
--- a/src/features/network-modifications/battery/modification/BatteryDialogTabsContent.tsx
+++ b/src/features/network-modifications/battery/modification/BatteryDialogTabsContent.tsx
@@ -6,7 +6,7 @@
  */
 
 import { Box, Grid2 as Grid } from '@mui/material';
-import { FormattedMessage } from 'react-intl';
+import { FormattedMessage, useIntl } from 'react-intl';
 import { BatteryDialogTab } from './batteryTabs.utils';
 import { GridSection } from '../../../../components/composite/grid/grid-section';
 import { GridItem } from '../../../../components/composite/grid/grid-item';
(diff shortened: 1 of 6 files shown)


src/features/network-modifications/battery/modification/BatteryDialogTabsContent.tsx
```diff
@@ -37,7 +39,29 @@ export function BatteryDialogTabsContent({
     voltageLevelOptions = [],
     PositionDiagramPane,
     fetchBusesOrBusbarSections,
+    fetchVoltageLevelEquipments,
 }: Readonly<BatteryDialogTabsContentProps>) {
+    const voltageRegulationField = (
+        <Box>
+            <SwitchInput name={FieldConstants.VOLTAGE_REGULATION} label="VoltageRegulationText" />
+        </Box>
+    );
```
Mathieu-Deharbe, line 63: This should use a `CheckboxNullableInput` and have a `previousValue` in order to handle the "not modified" state and display the current value in the network. Just like "Compensation" under it : 

Image

Mathieu-Deharbe (changes requested):
In the battery creation if voltageRegulatorOn is set to true, tension is mandatory : 

Image

But not in the modification : 

Image

And the application is fine without it : 

Image

Which one is right ? (the creation I guess?)

src/features/network-modifications/battery/modification/BatteryDialogTabsContent.tsx
```diff
@@ -79,6 +103,8 @@ export function BatteryDialogTabsContent({
                     </GridItem>
                 </Grid>
                 <Grid container spacing={2} paddingTop={2}>
+                    <GridItem size={4}>{voltageRegulationField}</GridItem>
+                    {voltageRegulationForm}
```
Mathieu-Deharbe, line 122: The alignement is wrong : 

Image

Mathieu-Deharbe (changes requested):
I can validate this form : 

Image

Which has distant regulation selected and no regulating terminal selected. And then, if I apply it : 

Image

(I am not sure that this has to be done, to be checked with the POs)

Mathieu-Deharbe (approved):

ayolab (approved):

PR #1270, opened 2026-08-03
move cspr form to commons-ui

anass45:
Moving the CSPR creation form, from gridstudy to commons-ui


src/features/network-modifications/static-var-compensator/creation/static-var-compensator-creation.types.ts
```diff
@@ -0,0 +1,42 @@
...
+    standbyAutomatonOn: boolean | null;
+    standby: boolean | null;
+    lowVoltageSetpoint: number | null;
+    highVoltageSetpoint: number | null;
+    lowVoltageThreshold: number | null;
+    highVoltageThreshold: number | null;
+    b0: number | null;
+    q0: number | null;
+    properties?: Property[] | null;
+}
```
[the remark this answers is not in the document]
anass45: Actually, this field isn't exposed in the frontend creation form—the backend handles setting this value automatically during creation

src/features/network-modifications/static-var-compensator/common/standby-automaton-form-utils.ts
```diff
@@ -0,0 +1,120 @@
...
+        is: true,
+        then: (s: Schema) => s.required(),
+    });
+
+const requiredWhenSusceptanceChoice = (schema: NumberSchema<number | null | undefined>) =>
+    schema.when([FieldConstants.ADD_STAND_BY_AUTOMATON, FieldConstants.CHARACTERISTICS_CHOICE_AUTOMATON], {
+        is: (addStandbyAutomaton: boolean, characteristicsChoiceAutomaton: string) =>
+            addStandbyAutomaton && characteristicsChoiceAutomaton === CHARACTERISTICS_CHOICES.SUSCEPTANCE.id,
+        then: (s) =>
+            s
```
KoloMenek: Maybe it could be better if the s was schema instead ? If you make the change here, it's worth having the same variable everywhere else.

src/features/network-modifications/static-var-compensator/common/set-points-limits-form-utils.ts
```diff
@@ -0,0 +1,151 @@
...
+        [FieldConstants.MAX_Q_AT_NOMINAL_V]:
+            nominalV != null && maxSusceptance != null ? maxSusceptance * nominalV ** 2 : maxQAtNominalV,
+        [FieldConstants.MIN_Q_AT_NOMINAL_V]:
+            nominalV != null && minSusceptance != null ? minSusceptance * nominalV ** 2 : minQAtNominalV,
+        [FieldConstants.VOLTAGE_SET_POINT]: voltageSetpoint,
+        [FieldConstants.REACTIVE_POWER_SET_POINT]: reactivePowerSetpoint,
+        [FieldConstants.VOLTAGE_REGULATION_TYPE]: voltageRegulationType,
+        [FieldConstants.VOLTAGE_LEVEL]: voltageLevelId ? { [FieldConstants.ID]: voltageLevelId } : null,
+        [FieldConstants.EQUIPMENT]:
+            equipmentId && equipmentType
```
KoloMenek, line 144: Is this necessary ? This wasn't in the original gridstudy form utils.

src/features/network-modifications/static-var-compensator/common/set-points-limits-form-utils.ts
```diff
@@ -0,0 +1,151 @@
...
+    voltageSetpoint: number | null;
+    reactivePowerSetpoint: number | null;
+    voltageRegulationType: string;
+    voltageLevelId: string | null;
+    equipmentType: string | null;
+    equipmentId: string | null;
+}) => ({
+    [FieldConstants.SETPOINTS_LIMITS]: {
+        [FieldConstants.CHARACTERISTICS_CHOICE]:
+            nominalV != null || minSusceptance == null
```
KoloMenek, line 129: Why the change from strict to loose comparaison ? Same comment for Line 136 and 138

src/features/network-modifications/static-var-compensator/common/set-points-limits-form-utils.ts
anass45: Yes it's necessary. In the original Gridstudy, the form was typed manually instead of being inferred from the Yup schema, which led to incorrect typing.

You can test it : if you remove the check on equipmentType, you will get the followig error in staticVarCompensatorDtoToForm : 
Type { id: string; type: string | null; } | null is not assignable to type { id: string; type: string; } | null

src/features/network-modifications/static-var-compensator/common/set-points-limits-form-utils.ts
anass45: In the old code, variables were typed as any, so there were no checks during calculations. Now that they are properly typed, you cannot pass undefined values to a squared exponent.

In fact, we only want to run the calculation if nominalV and minSusceptance are neither undefined nor null. Otherwise, we fall back to the default value minQAtNominalV.

So, if we aren't performing the calculation, for CHARACTERISTICS_CHOICE, we get Q_AT_NOMINAL_V; otherwise, we get SUSCEPTANCE.

You can test it by typing maxSusceptance, minSusceptance, and nominalV in the old code.

src/features/network-modifications/static-var-compensator/common/standby-automaton-form-utils.ts
anass45: done

src/features/network-modifications/static-var-compensator/creation/static-var-compensator-creation-form.tsx
```diff
@@ -0,0 +1,56 @@
...
+    PositionDiagramPane,
+    fetchBusesOrBusbarSections,
+    fetchVoltageLevelEquipments,
+}: StaticVarCompensatorCreationFormProps) {
+    const { tabIndex, setTabIndex, tabIndexesWithError } = useTabsWithError<StaticVarCompensatorDialogTab>(
+        STATIC_VAR_COMPENSATOR_TAB_FIELDS,
+        StaticVarCompensatorDialogTab.CONNECTIVITY_TAB
+    );
+    return (
+        <Stack spacing={2}>
```
dbraquart, line 34: can you apply here the same Stack/Box pattern than in LineForm plz ?

this will help for scrolling later (there is a pending work for this)

anass45: done

dbraquart: great

anass45:
PRs associées:
https://github.com/gridsuite/gridstudy-app/pull/4125
https://github.com/gridsuite/gridexplore-app/pull/938

